PR #3, opened 2017-11-30
Hardware: added support for KORAD KD3005P

zarath:
I've added the device identifier for my kd3005P.

It's electrical capabilities are identical to the ka3005p.


uwehermann:
Merged with a few tiny cosmetics, thanks!

PR #6, opened 2018-01-24
Unbreak libsigrok fx2lafw on FreeBSD

smortex:
These patches where sent to the sigrok-devel@ mailing-list but I got no feedback. Since sending patches over a mailing list is a good way to forget about them, and because pull-requests to this repository seems to be merged, please allow me to submit it again :wink:. Thanks!

-----------

After updating sigrok and pulseview on FreeBSD, I could not use my fx2lafw device anymore.  After investigating the issue, I found out the culprit: when `usb_get_port_path()` is called while the device is open, the function fails, but since the return value is never tested, the uninitialized memory that was supposed to receive the path is copied and considered valid, and execution continues.                                                                                                   

As a consequence, when attempting to compare the device path and uninitialized memory later on, no match is found.

Here are two commits, the first one ensure the device is closed before calling `usb_get_port_path()` (I could only test with the fx2lafw driver, but some drivers had the wrong order and where changed too).  The second patch test the return value of usb_get_port_path() so that errors are not ignored anymore.

### DIFF
--- a/src/device.c
+++ b/src/device.c
@@ -753,7 +753,9 @@ SR_API const char *sr_dev_inst_connid_get(const struct sr_dev_inst *sdi)
 			if (b != usb->bus || a != usb->address)
 				continue;
 
-			usb_get_port_path(devlist[i], connection_id, sizeof(connection_id));
+			if (usb_get_port_path(devlist[i], connection_id, sizeof(connection_id)) < 0)
+				continue;
+
 			((struct sr_dev_inst *)sdi)->connection_id = g_strdup(connection_id);
 			break;
 		}
--- a/src/hardware/chronovu-la/api.c
+++ b/src/hardware/chronovu-la/api.c
@@ -192,9 +192,9 @@ static GSList *scan(struct sr_dev_driver *di, GSList *options)
 			continue;
 		}
 
-		usb_get_port_path(devlist[i], connection_id, sizeof(connection_id));
-
 		libusb_close(hdl);
+
+		usb_get_port_path(devlist[i], connection_id, sizeof(connection_id));
 
 		if (!strcmp(product, "ChronoVu LA8"))
 			model = 0;
--- a/src/hardware/chronovu-la/api.c
+++ b/src/hardware/chronovu-la/api.c
@@ -194,7 +194,8 @@ static GSList *scan(struct sr_dev_driver *di, GSList *options)
 
 		libusb_close(hdl);
 
-		usb_get_port_path(devlist[i], connection_id, sizeof(connection_id));
+		if (usb_get_port_path(devlist[i], connection_id, sizeof(connection_id)) < 0)
+			continue;
 
 		if (!strcmp(product, "ChronoVu LA8"))
 			model = 0;
--- a/src/hardware/dreamsourcelab-dslogic/api.c
+++ b/src/hardware/dreamsourcelab-dslogic/api.c
@@ -214,9 +214,9 @@ static GSList *scan(struct sr_dev_driver *di, GSList *options)
 			continue;
 		}
 
-		usb_get_port_path(devlist[i], connection_id, sizeof(connection_id));
-
 		libusb_close(hdl);
+
+		usb_get_port_path(devlist[i], connection_id, sizeof(connection_id));
 
 		prof = NULL;
 		for (j = 0; supported_device[j].vid; j++) {
--- a/src/hardware/dreamsourcelab-dslogic/api.c
+++ b/src/hardware/dreamsourcelab-dslogic/api.c
@@ -216,7 +216,8 @@ static GSList *scan(struct sr_dev_driver *di, GSList *options)
 
 		libusb_close(hdl);
 
-		usb_get_port_path(devlist[i], connection_id, sizeof(connection_id));
+		if (usb_get_port_path(devlist[i], connection_id, sizeof(connection_id)) < 0)
+			continue;
 
 		prof = NULL;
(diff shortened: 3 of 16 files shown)


uwehermann:
Hi, thanks for your patch (and sorry for the delay)! The code looks good so far (will also test in a bit, but others have reported the fix works already, so it should be fine).

Maybe the first patch can be done a bit differently, though -- instead of forcing the driver authors to know/remember to use the correct order, could usb_get_port_path() simply check if the device is already open and simply close it before opening (or not re-open it twice)? If that's possible, all drivers (current and future ones) would work without any changes required and without things driver writers need to remember, which would be great.

I haven't actually looked into whether that's easily possible, just a random idea.

smortex:
I do not have much knowledge about the way to do things with USB, after a quick look at the FreeBSD code of libusb, I found that the older API (libusb-0.1) specifically [checked for an already opened device](https://svnweb.freebsd.org/base/head/lib/libusb/libusb01.c?revision=326219&view=markup#l160) (maybe this is not done in newer versions on purpose).

I do not see something in the public API to check if the device has already been open.  I asked for insights on IRC, maybe we will have valuable feedback from some FreeBSD developer :stuck_out_tongue_winking_eye:

hselasky:
Hi, It is possible that FreeBSD can pull out this information while enumerating the devices. Then it does not matter when it is called.
--HPS

smortex:
@hselasky that would be awesome I think!  In the meantime, do you have a short-term recommendation / best practice recommendation for this particular issue?

hselasky:
I think it would be best to have the libusb device handle open, before querying this information. I don't see how this can hurt.

uwehermann:
OK, so I've finally tried the patches on a FreeBSD 11.1 install on an amd64 laptop. I can confirm that without the patches you cannot get any samples from an fx2lafw device. After the patches the logs look different, but I'm still unable to get any samples.

Test: sigrok-cli -d fx2lafw -c samplerate=1mhz --samples 10 -l 5

I've also tried running sigrok-cli via sudo to be sure that it's not permissions related, but maybe that's not enough and/or I'm missing some other required steps on FreeBSD?

I'm seeing "Waiting for device to reset", "Unable to get version info: LIBUSB_ERROR_NO_DEVICE" and "Device failed to renumerate" here. Can anybody else reproduce this and/or suggest any fixes for this?

Thanks!

uwehermann:
The two patches are merged upstream now (with one additional fix in the new hantek-4032l driver), and mentioning the upstream Bugzilla bug #1109 (which is now marked as fixed). Thanks a lot!

I haven't yet been able to test this myself, still getting "Device failed to renumerate" for now, but I'll look into that and/or try installing FreeBSD on another box to see if it's a hardware problem or such.

I'm closing this issue as fixed, but if anyone wants to make an additional patch to move the fix into the common usb_get_port_path() function in usb.c, please go ahead. That would make any potential new drivers work automatically, without authors having to remember to use the correct ordering of the function calls.

smortex:
Thanks @uwehermann !

FYI, I bought a fx2lafw device to @hselasky who should receive it in a few weeks…  This will help if we need further tweaking as he is the main author of FreeBSD's USB stack 👍

hselasky:
Hi,

I've now investigated this issue and made the following patch for libsigrok:
[libsigrok.txt](https://github.com/sigrokproject/libsigrok/files/1841009/libsigrok.txt)

In addition there is a patch for FreeBSD's libusb:
https://svnweb.freebsd.org/changeset/base/331419

--HPS

hselasky:
Please note that under FreeBSD the same USB device can only be opened one time. The current libsigrok code tried to open the same USB device handle multiple times, due to the port path patch, and this won't work.

avg-I:
@hselasky, has the FreeBSD port / package been fixed?

smortex:
@avg-I regarding the port itself, there was no release of libsigrok and the fix was not added as a local patch, to if you are using one of the concerned devices, it will probably not work out of the box with the FreeBSD packages.

The change @hselasky did improves how things work at the OS level, but AFAICR, the fix of this PR is needed in all cases.  Maybe it's worth adding a patch to the ports tree.  @avg-I if you have problems with a driver, can you confirm that this patch fixes the problem?

avg-I:
@smortex, this is what I have with a vanilla FreeBSD package:
```
sigrok-cli --driver fx2lafw --show
Driver functions:
    Logic analyzer
Scan options:
    conn
fx2lafw:conn=2.2 - Saleae Logic with 8 channels: D0 D1 D2 D3 D4 D5 D6 D7
sr: fx2lafw: Unable to open device.
Failed to open device.
```

If I clone this repo and manually build libsigrok, then everything is okay for me:
```
LD_PRELOAD=libsigrok/.libs/libsigrok.so.4.0.0 sigrok-cli --driver fx2lafw --show
Driver functions:
    Logic analyzer
Scan options:
    conn
fx2lafw:conn=2.2 - Saleae Logic with 8 channels: D0 D1 D2 D3 D4 D5 D6 D7
Channel groups:
    Logic: channels D0 D1 D2 D3 D4 D5 D6 D7
Supported configuration options across all channel groups:
    continuous: on, off
    limit_samples: 0 (current)
    conn: 2.2 (current)
    samplerate - supported samplerates:
      20 kHz
...
      24 MHz
    Supported triggers: 0 1 r f e 
    captureratio: 0 (current)
```
I haven't tried `libsigrok.txt` from @hselasky

smortex:
@avg-I I opened a PR with the local diff of my package:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=228789

PR #14, opened 2018-10-19
Python-Binding: Added data array for logic packet payload

littleyoda:


bindings/python/sigrok/core/classes.i
```diff
@@ -548,4 +548,23 @@ std::map<std::string, Glib::VariantBase> dict_to_map_options(PyObject *dict,
 }
 }
 
+/* Return NumPy array from Analog::data(). */
```
martinling: Copy-paste mistake I think - comment should say Logic::data.

bindings/python/sigrok/core/classes.i
```diff
@@ -548,4 +548,23 @@ std::map<std::string, Glib::VariantBase> dict_to_map_options(PyObject *dict,
...
+%extend sigrok::Logic
+{
+    PyObject * _data()
+    {
+        npy_intp dims[1];
+        dims[0] = $self->data_length();
+        int typenum = NPY_UINT8;
```
martinling: It might be better to look at the unitsize of the logic data and return an array of the appropriate type/dimensions.
E.g. if using a 16-channel device with unitsize 2 returning 100 samples, that would give you a length 100 uint16 array rather than a length 200 uint8 array.

littleyoda: i've also thought about this approach, but what happens with really big logic analyser? i don't know if uint128 or even unit256 is supported.

martinling: Yeah, that's a good point.
I think the best solution then is to return a uint8 array of shape (N, M) where N is the number of samples and M is the logic data unitsize in bytes.

littleyoda: done

bindings/python/sigrok/core/classes.i
```diff
@@ -548,16 +548,17 @@ std::map<std::string, Glib::VariantBase> dict_to_map_options(PyObject *dict,
...
-/* Return NumPy array from Analog::data(). */
+/* Return NumPy array from Logic::data(). */
 %extend sigrok::Logic
 {
     PyObject * _data()
     {
-        npy_intp dims[1];
-        dims[0] = $self->data_length();
+        npy_intp dims[2];
+        dims[0] = $self->data_length() / $self->unit_size();;
```
martinling: You've got an extra semicolon on the end here.

PR #16, opened 2018-10-31
Fix mismatched printf specifiers in beaglelogic_tcp.c

janisozaur:


uwehermann:
Did you see compiler warnings for these, if yes on which compiler, which version, which OS etc?

Anyway, please update to use the standard macros like PRIu32 for uint32_t, etc.

uwehermann:
Squashed in the change to use PRIu32 for every uint32_t and merged in e3f86ef5fc4e5061a19b4af72727077797446fbd, thanks!

janisozaur:
Took a while

PR #18, opened 2018-12-02
Add support for YiXingDianZi MDSO in hantek-6xxx driver.

majekw:
Basic support for YiXingDianZi (Noname/OEM) MDSO.
It works, but V/div on '10' range is incorrect and fixing this require some rewrite of whole hantek-6xxx driver.


uwehermann:
Merged, thanks a lot!

PR #19, opened 2018-12-04
Check if limit samples is valid (less or equal the size of the ringuffer)

danselmi:


uwehermann:
Merged in f3549a1ccf7bacdc13827e58c96912da7c631a67 with one small whitespace-fix (and added the "ipdbg-la: " prefix in the commit message), thanks!

PR #20, opened 2018-12-05
scpi-dmm: add experimental Keysight 34465A support

mikebwilliams:
Tested DC current, DC voltage, and resistance. Instrument gives an error
only on first measurement but the output of the measurements is correct.


uwehermann:
Perfect! Merged, thanks a lot!

PR #22, opened 2019-01-18
Fix IPDBG LA driver on windows

danselmi:
The IPDBG LA driver is not working on windows.
First patch is working on windows but unusable slow.
Second patch improves speed.
Tested on linux fedora 29 and windows 7


uwehermann:
Merged with a few minor cosmetic/comment fixes, thanks!

PR #24, opened 2019-03-17
korad-kaxxxxp: support for Stamos S-LS-31 power supply

andresth:
This is another rebranded Korad device.
See http://www.stamos-welding.com/mains-adapter-s-ls-31 for specs


uwehermann:
Merged in 27a9b6639e1bd6530bf8f4304677dc127e343001, thanks!

PR #26, opened 2019-05-13
warn only if salea fpga (clone) is unsupported

pieska:
don't exit with error if the salea fpga is detected as unsupported. just issue a warning with the detected version and continue. i have such a clone and it works with the original salea software and with sigrok despite the fact, it's FPGA version is 0xff.


uwehermann:
Merged with some rewording of the commit message in 108d2ec2c23df2dab1923039fbd1e1a6b2aa7c9b, thanks!

PR #29, opened 2019-07-20
Add Korad KD6005P

tachiniererin:
This adds the Korad KD6005P power supply which can source up to 60V and 5A on its output.


uwehermann:
Merged in b1b8a7d079606a095faa582d8de78663f749384e (with added "korad-kaxxxxp:" prefix in the commit message), thanks a lot!

PR #30, opened 2019-08-21
rigol-ds: Add initial Rigol MSO5000 support.

sre:
This adds basic support for the Rigol MSO5000 series. It has
the same problems as the DS4000 series: Live capture provides
one digital channel per byte. Buffered memory returns the data
compressed (one byte has 8 digital channels), but the banks are
read separately. It's not possible to read uint16.


sre:
I brought my MSO5072 to CCCamp2019, feel free to visit the Mainframe village:

https://map.events.ccc.de/#19/53.03146/13.30686

uwehermann:
Look great. Merged in f6129c8f0c92e45de5d70b4101bf2bd759a5fdf7, thanks!

PR #31, opened 2019-09-16
std: Remove device closure in acquisition stop

abdulkerim-altuntas:
In the standard API helper function std_serial_dev_acquisition_stop()
the device is closed before returning which is actually unnecessary
and conflicts with the implementation of other device drivers that have
their own acquisition stop function.

### DIFF
--- a/src/std.c
+++ b/src/std.c
@@ -392,11 +392,6 @@ SR_PRIV int std_serial_dev_acquisition_stop(struct sr_dev_inst *sdi)
 		return ret;
 	}
 
-	if ((ret = sr_dev_close(sdi)) < 0) {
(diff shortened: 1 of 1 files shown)


uwehermann:
Thanks for the patch! There was actually a bug report for this one in Bugzilla (bug #1271) for a while now, but I somehow missed that it also had a patch attached. Merged that one now (cfdc80151b106ae83f3380cb308d53314974257a), since it was the first one. But they're identical anyway, so the issue should be fixed.

abdulkerim-altuntas:
Alright, thanks for the update!

Kind regards,
Abdulkerim Altuntas

PR #32, opened 2019-10-01
Initial support for MASTECH MS6514 thermometer

DaveBuechi:
Add support for MASTECH MS6514 dual channel thermometer, see https://sigrok.org/wiki/MASTECH_MS6514


uwehermann:
Merged in 23669c3df33a3ea884afabf4b7a8cd6fe4a75824 with a few minor whitespace changes squashed in, plus some commits on top, thanks a lot!

Thanks for the wiki page as well! If possible, please add the missing "lsusb -v" (as root) there, that's always nice to have as a reference.

PR #33, opened 2019-10-30
[lecroy-xstream] Remove header read

sylvainpelissier:
The header read prevents to acquire a waveform from oscilloscope.

### DIFF
--- a/src/hardware/lecroy-xstream/protocol.c
+++ b/src/hardware/lecroy-xstream/protocol.c
@@ -645,12 +645,6 @@ SR_PRIV int lecroy_xstream_receive_data(int fd, int revents, void *cb_data)
 	if (ch->type != SR_CHANNEL_ANALOG)
 		return SR_ERR;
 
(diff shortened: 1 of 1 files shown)


abraxa:
Which model did you connect to?

sylvainpelissier:
The WaveRunner 625ZI. Without the header read it works fine.

uwehermann:
Hm, wouldn't this change break other models which have that header?

sylvainpelissier:
To my understanding according to the [manual](http://cdn.teledynelecroy.com/files/manuals/wm-rcm-e_rev_d.pdf) we used the command **COMM_HEADER OFF** in the function lecroy_xstream_init_device. Thus the oscilloscope should not send a header.

uwehermann:
Sounds good, yeah. Merged (with minor commit msg changes and dropping the now-unused "buf" variable) in 24931412eeee31e5c0bb5291a9c73e7b57309216, thanks!

PR #35, opened 2019-11-17
Add support for RND 320-KD3005P

astrand:


uwehermann:
Merged (with some minor whitespace and commit msg changes squashed in) in cf6beeb20d46188f33eb5fe3ccee927d8aa92003, thanks!

astrand:
At least on my system, sigrok-cli with this device is not 100% stable. Sometimes it fails with:

sr: korad-kaxxxxp: Unknown model ID '8D 320-KD3005P V4.2' detected, aborting.

and similar errors. Another problem is that it's quite slow to startup.

PR #36, opened 2019-12-03
Added minimal support for HP34401A to scpi-dmm

artgodwin:
This could do with some more work as the commands are quite slow and will block.
I want to add some keithley DMMs as well and don't want to do a more complete implementation until I have worked with them.


uwehermann:
Merged (with some cosmetics squashed in) in a16198316fc68440783ad1040ac8ddd9b918e67b, thanks!

PR #38, opened 2020-02-08
Add UNI-T UT804 DMM definitions for serial and usb

tachiniererin:
The UT804 has basically the same hardware than the UT71x series, just a larger display and huge mostly empty plastic shell. Protocol is the same as described in the wiki and matches what UNI-T claims it talks: https://www.uni-trend.com.cn/uploadfile/2019/1021/20191021050157366.pdf

This was tested on the bench with a UT804 and a Raspberry Pi.


uwehermann:
Merged in ca7d442692a1785129cb5caba57c280bf5ce2a9f (with minor commit message cosmetics), thanks a lot!

Feel free to add a wiki page for the device btw, IIIRC you should already have a wiki account.

PR #39, opened 2020-02-14
Enable Edge-Triggering capabilities on the Zeroplus Logic Analyzer 

jmA500:
This should enable Edge triggering capabilities on the Zeroplus Logic Analyzer family. This has been tested with only one device (LAPC-32128) but should work for the whole family.


uwehermann:
Merged in 88daa0536cb4107ac4d8ddaacedd06fa39d55cc0 with minor whitespace changes, thanks a lot!

I've also added the usual "zeroplus-logic-cube: " prefix in the commit message and mentioned the bug number of an existing bug we had for this:

https://sigrok.org/bugzilla/show_bug.cgi?id=1334

PR #40, opened 2020-02-15
rdtech-um: Add initial support for the RDTech UMxx series

andysan:
This changeset adds support for the RDTech UMxx series of USB power
meters. The driver has been tested with the RDTech UM24C, but should
support the UM24C, UM25C, and the UM34C.

Currently, the driver reports the following channels:
  * V: VBus voltage
  * I: VBus current
  * D+: D+ voltage
  * D-: D- voltage
  * Temp: Device temperature
  * Consumption: Energy consumed in threshold-based recording mode.

The number of significant digits shown for each channel has been set
to match the number of digits shown on a UM24C.

Missing features:

  * There is currently no support for configuring threshold-based
    recording from Sigrok, but this can be done on the device itself.

  * Fast charging mode currently not logged.

Usage example:

sigrok-cli -d rdtech-um:conn=bt/rfcomm/MAC --scan
sigrok-cli -d rdtech-um:conn=/dev/rfcomm0 --scan

Known issues:

  * When using Sigrok's Bluetooth transport implementation, the device
    is disconnected between probing and sampling. Some devices (e.g.,
    the UM24C), dislikes this and can't be reconnected reliably for
    sampling. This is not an issue when setting up a rfcomm device
    manually and using it as a serial port.

Kudos to Sven Slootweg for documenting most of the protocol.

Signed-off-by: Andreas Sandberg <andreas@sandberg.pp.se>


src/hardware/rdtech-um/protocol.c
```diff
@@ -0,0 +1,281 @@
...
+#define UM_CMD_POLL 0xf0
+
+static const struct rdtech_um_channel rdtech_default_channels[] = {
+	{ "V",  2, RDTECH_UM_UINT16, 0.01, 2, SR_MQ_VOLTAGE, SR_UNIT_VOLT },
+	{ "I",  4, RDTECH_UM_UINT16, 0.001, 3, SR_MQ_CURRENT, SR_UNIT_AMPERE },
+	{ "D+", 96, RDTECH_UM_UINT16, 0.01, 2, SR_MQ_VOLTAGE, SR_UNIT_VOLT },
+	{ "D-", 98, RDTECH_UM_UINT16, 0.01, 2, SR_MQ_VOLTAGE, SR_UNIT_VOLT },
+	{ "Temp", 10, RDTECH_UM_UINT16, 1.0, 0, SR_MQ_TEMPERATURE, SR_UNIT_CELSIUS },
+	/* Threshold-based recording (mWh) */
+	{ "Consumption", 106, RDTECH_UM_UINT32, 0.001, 3, 0, SR_UNIT_WATT_HOUR },
```
knarfS: I'd suggest to rename the temperature channel to `T`, like in other devices (e.g. in the `appa-55ii` driver).
The consumption channel should be renamed to `Wh` or `E` for energy (or `W` for work, but `W` would be ambiguous to Watt). I'd prefer `Wh` :)

Also the measured quantity is missing here. I mean there is no `SR_MQ_ENERGY` (or `SR_MQ_WORK`) in the `sr_mq` enum (`include/libsigrok/libsigrok.h`). Would be great, when you would add the missing mq to the enum (at the end of the enum to maintain API/ABI compability), to the `sr_key_info_mq` struct in `src/hwdriver.c` and to `bindings/cxx/enums.cpp`.

Same for the `rdtech_um25c_channels` down below.

src/hardware/rdtech-um/api.c
```diff
@@ -0,0 +1,175 @@
...
+
+#define RDTECH_UM_SERIALCOMM "115200/8n1"
+
+static const uint32_t scanopts[] = {
+	SR_CONF_CONN,
+	SR_CONF_SERIALCOMM,
+};
+
+static const uint32_t drvopts[] = {
+	SR_CONF_MULTIMETER,
```
knarfS: There is a config key (device type) `SR_CONF_ENERGYMETER` which is perhaps a better fit for this device?

andysan: Would it be appropriate to list it as both `SR_CONF_MULTIMETER` and `SR_CONF_ENERGYMETER`?  I'm not entirely sure how this is used by the library users.

src/hardware/rdtech-um/protocol.c
andysan: Thanks for the review! I think I prefer `E` to `Wh` for the measured energy consumption. I have updated the commit accordingly.

knarfS: Yeah, `E` or `Wh` doesn't really matter. I'm using `Wh` for detecting and naming energy channels in SmuView, but I'm going to add `E` to the detection part.

You missed `bindings/cxx/enums.cpp` when adding `SR_MQ_ENERGY`:

- At line 116:

        const Quantity Quantity::_ENERGY = Quantity(SR_MQ_ENERGY, "ENERGY");
- At line 149:

        const Quantity * const Quantity::ENERGY = &Quantity::_ENERGY;
- At line 183:

        {SR_MQ_ENERGY, Quantity::ENERGY},

src/hardware/rdtech-um/api.c
knarfS: There are devices that have more than one device type (like scopes with digital channels), but technically speaking this isn't a multimeter with switchable measurement types.

Only place I'm aware of where the device type is evaluated, is in SmuView. :)
It is f.e. used to show different GUI elements for different device types and to auto generate math channels for power supplies and loads.
Though at the moment, multimeters and power meters are both handled as "measurement devices". So no special treatment for power meters, right now.

src/hardware/rdtech-um/protocol.c
andysan: Isn't that file auto-generated? I did some grepping in the repo earlier to make sure I didn't miss anything and it didn't show up.

knarfS: Ups, I'm sorry. Of course, you are right!

src/hardware/rdtech-um/api.c
andysan: So maybe it should be listed as a combination of `SR_CONF_POWERMETER` and `SR_CONF_ENERGYMETER` in that case? I'm honestly not sure why power meters and energy meters have their own types, it seems like most energy meters would also measure power and as far as I can tell, there is exactly one device of either type in Sigrok at the moment (both are listed as energy meters on the wiki though).

andysan:
Ping! This has been sitting here since February.

uwehermann:
Looks good, thanks a lot! Sorry for the delay. Merged in 8b607a24e7e7dec8b65963281b9cb43a0c5ba0a6 and a few commits before that.

I squashed some minor cosmetics/whitespace changes (e.g. "Sigrok" -> "sigrok" and other stuff) and added one commit on top.

Will test on hardware over the weekend or so, I have UM24C/25C/34C around.

PR #41, opened 2020-02-15
fluke-dmm: Cleanup Fluke 28x QM handling

andysan:
The code that parses QM responses from Fluke 28x DMMs uses a long series of if-else statements to determine the type of measurement reported by the device. Rewrite this to use a conversion table that contains a mapping between DMM units and Sigrok's internal units.

This change is a first step towards using QDDA to query the DMM and get additional information such as significant digits.

PR #43, opened 2020-02-24
RFC: Support multiple channels in Fluke 28x DMMs

andysan:
This is a pull request to get early feedback on a set of changes that add support for logging all of the channels (live, primary, secondary, etc.) reported by the Fluke 28x series of DMMs. The channel names in this pull request are not final and currently only reflect the names in the protocol specification.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -313,6 +313,7 @@ if HW_FLUKE_DMM
 src_libdrivers_la_SOURCES += \
 	src/hardware/fluke-dmm/protocol.h \
 	src/hardware/fluke-dmm/protocol.c \
+	src/hardware/fluke-dmm/fluke-28x.c \
 	src/hardware/fluke-dmm/api.c
 endif
 if HW_FTDI_LA
--- a/src/hardware/fluke-dmm/api.c
+++ b/src/hardware/fluke-dmm/api.c
@@ -53,11 +53,11 @@ static const char *scan_conn[] = {
 };
 
 static const struct flukedmm_profile supported_flukedmm[] = {
-	{ FLUKE_187, "187", 100, 1000 },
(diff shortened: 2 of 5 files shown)


andysan:
Ping @knarfS. You seem like someone who might have opinions about this. :)

andysan:
I think the first three commits should be OK for mainline. I'll deal with the first one in #41 and then revisit this one after that. I think it would make sense to merge the first three (it would end up being the first two after a rebase) changes separately. Would it be best if I push them as a separate PR?

abraxa:
@andysan With #41 being merged, I'd be happy to revisit this PR once you rebased the changes on top of master.

andysan:
I have rebased this now, but I would like to test it a bit more before it is merged. The two first commits should be uncontroversial and could be merged as is though. The main improvement from the third commit is that it is the foundation for multi-channel support and that it adds a bit more metadata to readings. The main benefit from multichannel support is that it adds support for modes where the instrument simultaneously provides multiple readings (e.g., separate AC and DC components).

There are a couple of things I would like to fix in the last commit. I'm not happy with the way channel names are far from the enum that it depends on, this seems quite fragile. I think the right solution might be to just use a variant-specific helper function that sets up the channels if a specific variant supports multiple channels.

biot:
First patches are again a very nice refactoring, no comments.

What jumps out at me about the WIP is that you want to create multiple channels because you're getting multiple MQs -- but it's actually common to just send this sort of thing out from the same channel, with different MQs instead, that's what they're for.

For example, the programmable power supplies typically send out a continuous stream of the voltage and current they're outputting. The driver often has to query for those separately, even. But they're all sent from the same channel, not "CH1-voltage" and "CH1-current" or something.

PR #44, opened 2020-02-24
bt/bt_bluez: Implement retry if rfcomm sockets are busy

andysan:
There are cases where the connect() call returns EBUSY when trying to
connect to a device. This has been observed when sampling an RDTech
UM24C. In this case, scanning the device works fine. However, when
sampling the device, Sigrok first scans the device, then closes the
connection and re-opens it to sample the device. If the close/open
calls happen in close successions, the Bluetooth stack sometimes
returns EBUSY.

Work around this issue by retrying if the connect() returns EBUSY.

Signed-off-by: Andreas Sandberg <andreas@sandberg.pp.se>

### DIFF
--- a/src/bt/bt_bluez.c
+++ b/src/bt/bt_bluez.c
@@ -95,6 +95,9 @@
 #define STORE_MAC_REVERSE	1
 #define ACCEPT_NONSEP_MAC	1
 
+#define CONNECT_RFCOMM_TRIES	3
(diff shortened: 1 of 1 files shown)


uwehermann:
Merged with minimal whitespace changes in 4704f64551dd917c2616b4162e7e816fb57113da, thanks!

PR #45, opened 2020-02-24
input/trace32_ad: Fix pod_data uninitialised warning

andysan:
There are code paths where the 'pod_data' variable can be used without being initialised. Fix this to avoid a compiler warning.


uwehermann:
Merged in 35037b1d8b6019c3605f6b579022ce35aefdb759, thanks!

PR #46, opened 2020-03-06
Add initial support for the RDTech TC66C

andysan:
These changes add support for the RDTech TC66C USB power meter.

Currently, the driver reports the following channels:
- V: VBus voltage
- I: VBus current
- D+: D+ voltage
- D-: D- voltage
- E: Energy consumed in threshold-based recording mode.

The number of significant digits shown for each channel has been set to match the number of digits shown on the device.

Usage example:
`sigrok-cli -d rdtech-tc:conn=/dev/ttyACM0 --scan`

Known issues:
 - BLE support is currently unimplemented. This uses a different command set, but the same poll data format.

Kudos to Ben V. Brown for reverse engineering some of the protocol and documenting the encryption key used for poll data. See https://sigrok.org/w/index.php?title=RDTech_TC66C for the protocol specification.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -50,6 +50,7 @@ lib_LTLIBRARIES = libsigrok.la
 # Backend files
 libsigrok_la_SOURCES = \
 	src/backend.c \
+	src/binary_helpers.c \
 	src/conversion.c \
 	src/crc.c \
 	src/device.c \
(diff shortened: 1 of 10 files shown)


uwehermann:
Looks good, thanks a lot! Merged in cae33a58743e408a602771d6924ee8c271326f47 and a few commits before that.

I squashed some minor cosmetics/whitespace changes and added the missing crc.c as well.

Will test on hardware over the weekend or so, I have a TC66C around.

The new dependency on nettle is a bit unfortunate (though I've kept it for now). Maybe we can juts drop a simple single-c-file open-source AES impementation into libsigrok src/ to avoid increasing the dependencies, not sure.

PR #48, opened 2020-03-25
gwinstek-gpd: Fix use of uninitialized variable

andysan:
The variable 'ret' is sometimes uninitialized when returned from config_get.  Fix this behavior.


uwehermann:
Merged in 5ff772410b732c650f281a55dd839d96d0f80ab1, thanks!

PR #49, opened 2020-03-30
Include missing header file

jmbr:
Including the `functional` header is necessary in order to use `std::function` in the definition of `LogCallbackFunction` when compiling with gcc 7.3.0.


uwehermann:
Merged in 5652085a4ee8884c913ec0d9c254fcd0a432df66, thanks!

I added "bindings/cxx: " as prefix in the commit message to make git history browsing a little easier.

PR #52, opened 2020-04-02
Improve ASCII output

v1ne:
* Align channels, even if their names have different lengths.
* Print the trigger if less samples than width are captured.

### DIFF
--- a/src/output/ascii.c
+++ b/src/output/ascii.c
@@ -39,7 +39,6 @@
 struct context {
 	unsigned int num_enabled_channels;
 	int spl;
-	int bit_cnt;
 	int spl_cnt;
 	int trigger;
 	uint64_t samplerate;
--- a/src/output/ascii.c
+++ b/src/output/ascii.c
@@ -154,6 +154,23 @@ static GString *gen_header(const struct sr_output *o)
 	return header;
 }
 
+static void maybe_add_trigger(struct context *ctx, GString *out) {
+	if (ctx->trigger <= -1)
+		return;
+
+	int offset = ctx->trigger;
+
+	/*
+	 * Sample data lines have one character per bit and
+	 * no separator between bytes. Align trigger marker
+	 * to this layout.
+	 */
+	g_string_append_printf(out, "%*sT:%*s^ %d\n", ctx->max_namelen - 1, "", offset, "", offset);
+
+	ctx->trigger = -1;
+}
+
+
 static int receive(const struct sr_output *o, const struct sr_datafeed_packet *packet,
 		GString **out)
 {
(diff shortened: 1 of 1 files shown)


v1ne:
@uwehermann: Is there anything I can help with?

gsigh:
I'd disagree with the "same length" part of this PR. Because umlauts (yes we've seen them in channel names). So the resulting output would be as arbitrarily unaligned as the current implementation would be.

v1ne:
@gsigh: Agreed, it's only the same length if all characters are encoded fixed-length, e.g. with ISO 8859-1 or ASCII. I can enhance the commit message to mention this. Thus, a solution for you could be to use ISO 8859-15 (including Euro sign).

For UTF-8, I assume that terminal emulators put one grapheme cluster into every one of these rectangles that used to contain a "character". Grapheme clusters can be counted by iteration, as defined in Unicode TR29.†
Just: Is there any Unicode support library available in this codebase that can cut a string into grapheme clusters? I don't want to pull in libICU or Boost.Locale for this tiny feature. If you provide me with an implementation of the TR29 grapheme cluster segmentation algorithm, I'd be happy to use it!
Do you see a simple way forward that gives us support for more-than-just-ASCII characters? I'd be happy to do that!

†:
If you want to know the gory details: It's impossible to do this correctly at this layer because it's up to the font shaper to turn a sequence of code points into grapheme clusters. Concretely, think of rendering Devanagari or emoji. The TR29 "Grapheme Cluster Boundary" rules are overridden by your font which support only certain code point sequences and has certain coalescing rules.  For emoji, think of the proprietary ninja cat (🐱‍👤): If you treat it as one grapheme, because your coalescing rules are lax, and your font doesn't support it, the font shaper will refuse to merge the constituent parts (🐱, ZWJ, 👤) into one grapheme, ending up with two graphemes (🐱👤). If you're strict with your coalescing rules, you might refuse the unofficial sequence, but the font shaper will recognise it. Both will ruin your layout.

gsigh:
That was the point I tried to make. I'm well aware of the complexity of encodings, the impossibility of getting a text's display width from naive byte counts, and how huge the dependencies are to get there. And I agree that this is way out of the scope of a library like libsigrok. The question is not whether _I_ can help myself, but what _users_ do with the software. And we cannot force them to run a specific setup.

So I agree that the other two parts of the PR are worth taking. But the channel text length would be as arbitrary after the change as it was before the change, and there is no simple solution to address that issue properly. So I suggest to not take this part.

v1ne:
Yeah, since unicode, console programs are between a rock and a hard place…

Yet, I still disagree that padding the names to the same byte count wouldn't help at all. It's not the final solution since it's still broken for most non-ASCII names — agreed. But solves the issue at least for pure-ASCII names. Thus, I'd consider it an incremental improvement. If we still disagree, I can also just drop the commit in question, sure.

gsigh:
Merged in cc835205cdc3 with some adjustment on top. Thank you!

PR #53, opened 2020-04-03
RDTech-DPS: Make it work again

v1ne:
I couldn't make this work with the serial connection (CH341). With these patches, the power supply works in SmuView and sigrok-cli.


uwehermann:
The mutex change is merged in 69c5d959e75b9bf9689be39da0dd8abd6078d89a, thanks!

The other issues should be fixed slightly differently. Generally each driver's scan() API call should close any open drivers or handles, later calls to open() will (re-)open the devices as needed.

Please send another PR with the changes to scan/open, thanks!

(also, make it one two commits please, one per driver)

PR #54, opened 2020-04-05
Add support for RND KA3005P V5.5 power supply

Feuerwerk:
I recently bought a RND KA3005P which identify itself as _RND 320-KA3005P V5.5_, so i added this id to the korad driver.


uwehermann:
Merged with added "korad-kaxxxxp: " commit message prefix in c8dcd3ab72ff9ced54d099849d67404b39edce83, thanks!

PR #55, opened 2020-04-06
Add support for TENMA 72-2540 V5.2 power supply

PetervdPerk:
I've purchased a TENMA 72-2540 power supply (Korad KAXXXP compatible), which reports "TENMA 72-2540 V5.2 SN:XXXXXXXX" as ID reply. This PR adds support for this new ID reply.


uwehermann:
Merged with added "korad-kaxxxxp: " commit message prefix in 082ca8d8bc3dc326cd0507a203dfedd0ea0b7de6, thanks!

PR #56, opened 2020-04-08
scpi-pps: fixed out-of-bounds array access

fastflo:
when accessing devc->device->channels in config_list().
this array has to be accessed via the "hw_output_idx".

found this invalid memory access when testing with a 4 channel PPS that can measure two output units per channel. config_list() would return garbage for channels 3 & 4.

### DIFF
--- a/src/hardware/scpi-pps/api.c
+++ b/src/hardware/scpi-pps/api.c
@@ -674,7 +674,7 @@ static int config_list(uint32_t key, GVariant **data,
 {
 	struct dev_context *devc;
 	struct sr_channel *ch;
-	struct pps_channel* pps_ch;
+	struct pps_channel *pch;
 	const struct channel_spec *ch_spec;
 	int i;
 	const char *s[16];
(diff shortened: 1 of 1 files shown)


fastflo:
also filed a bug report for this: https://sigrok.org/bugzilla/show_bug.cgi?id=1533

uwehermann:
Hi, thanks for the PR!

Generally when there's a PR, there's no need to necessarily also file a bug report; but it also doesn't hurt if you like to add one. However, if there's a Bugzilla item, please add "This fixes bug #1234." to the end of the commit message for cross-referencing.

In the commit, please squash a few cosmetic changes if possible:
* Please use "struct pps_channel *pch;" as name for consistency, that's already used elsewhere in the driver.
* Generally, code comments should start with capital letter and end with full stop in almost all cases (not related to this specific PR, just a general comment, for consistency). The same goes for commit messages ideally, though it's more of an optional thing there if you prefer all-lowercase for some reason there.

uwehermann:
Squashed and merged in 8f3c77db26dece82109b386ad5ba7ea344b1ec8f, thanks!

PR #57, opened 2020-04-08
device support for R&S HAMEG HMP4040 PPS via HO720

fastflo:
i tested for all 4 channels (with #56 applied):
- setting/getting target voltage & current limit
- measuring voltage & current
- controling & reading output enable
- setting/getting OVP limit
- reading OVP TRIP state
- reading CC or CV mode

untested/unsure: reading OTP state

linux reports that HO720 module as:
```
[31102.063593] usb 4-2.4.4: New USB device found, idVendor=0403, idProduct=ed72, bcdDevice= 4.00
[31102.063597] usb 4-2.4.4: New USB device strings: Mfr=1, Product=2, SerialNumber=3
[31102.063599] usb 4-2.4.4: Product: HAMEG HO720
[31102.063600] usb 4-2.4.4: Manufacturer: HAMEG
[31102.063602] usb 4-2.4.4: SerialNumber: .........
[31102.065546] ftdi_sio 4-2.4.4:1.0: FTDI USB Serial Device converter detected
[31102.065620] usb 4-2.4.4: Detected FT232BM
```


fastflo:
i just realized how to use the update_status-hook, i am going to send further commits...

fastflo:
> i just realized how to use the update_status-hook, i am going to send further commits...

no, from reading scpi_pps_receive_data() and e.g. hp_6630b_update_status() i don't understand how to send channel-specific OVP/OTP/REG changes down the session-bus. those update() methods are only called once after sampling voltage&current for all enabled channels...
maybe someone else can look into this or provide hints...

fastflo:
with this last commit i poll the status questionable register for each channel and generate meta packets for the session bus.
i also somewhat re-used/mis-used SR_CONF_DATA_SOURCE in a meta packet to signal consumers which hardware channel following meta packets should be assigned to.

fastflo:
with the last commits only "enabled" channels are queried for their ovp/otp/reg state.

knarfS:
> i also somewhat re-used/mis-used SR_CONF_DATA_SOURCE in a meta packet to signal consumers which hardware channel following meta packets should be assigned to.

In general, I like this quick fix for the missing configurable/channel group in the meta package. But I would suggest to add a new config key, f.e. `SR_CONF_CONFIGURABLE` or `SR_CONF_CONFIGURABLE` or even something more explicit.
The reason is, that `SR_CONF_DATA_SOURCE` is used in some drivers that have more that one channel group (like the `rigol-ds` and `siglent-sds`). Scopes are not supported by SmuView atm, but will be in the future and the misuse of `SR_CONF_DATA_SOURCE` could lead into some undefined behavior.

I haven't tested it yet, but will in the next two weeks.

For reference, this is my attempt to solve this issue in a clean way, but that require an API change and therefore is not mergeable yet:
https://github.com/knarfS/libsigrok/tree/v0.6.0 commit 80639b8 (_Change the API for sr_datafeed_meta by adding sr_channel_group._)

fastflo:
to avoid the confusion with other devices use of DATA_SOURCE i did add a new SR_CONF_CHANNEL_GROUP for now.
i understand that this might not be the optimal solution and your API change is to be preferred.

gsigh:
I cannot comment on the extension of the meta packet. But I could agree with the introduction of the HMP support. Can you shuffle the set such that introduction of support for the PSU goes first and the optional status update then follows? This would let me pick up the first half, until the other half got shaken out.

gsigh:
Basic support for the HMP4000 series went mainline. Which leaves the status update and per-channel details where I cannot help.

gsigh:
Want to update your tree after part of the set was taken?

fastflo:
> Want to update your tree after part of the set was taken?

yes, i will as soon as i find time. i hope within the next 7 days.

PR #58, opened 2020-04-08
Modbus: Close device after scan

v1ne:
Following up in #53, this change unbreaks my RDTech DPS and should fix a similar issue for the Maynuo M97. Since `sr_modbus_probe` opens the underlying device, I think it can also close it without mentioning any of this in the documentation.


uwehermann:
Merged with minimal whitespace changes in 1d657f47be01b2194f6caa7c55d95fc869bd86dc, thanks!

PR #60, opened 2020-04-11
device support for Kingst LA2016 USB logic analyzer

fastflo:
i also wrote about this on sigrok-devel on 2020-03-21 and didn't receive any response, so i'll try once again, maybe this microsoft owned tool is preferred.

i am happy to receive any requests/suggestions for changes.


uwehermann:
Merged in f2cd2debf9dcbf0e83ec70d8ea41d8421a400dfd with various minor whitespace/cosmetic fixes squashed in, thanks a lot!

Added two commits (ARRAY_SIZE and udev rule) on top.

Did a quick test on my device with this (77a1:01a2) using the firmware extracted from the current vendor software (3.4.2), but that seems to not work at the moment, acquisitions seem to "hang".

Added more info here: https://sigrok.org/bugzilla/show_bug.cgi?id=1559

Can you reproduce this? Probably some differences in recent vendor files(?)

fastflo:
i tested with vendors linux software version 3.4.0 md5 ca407133cb83b700983d2b704a4255c2 and the same usb vendor/product id as yours..

i will look into this.

PR #63, opened 2020-05-11
Fix CSV time by avoiding accumulating errors

earlephilhower:
As reported in https://sigrok.org/bugzilla/show_bug.cgi?id=1027 , the
CSV reported time in ns can have an error that increases over time when
the sample frequency does not divide into an integer number of
nanoseconds (i.e. the common 24MHz USB analyzers).

Fix the error by calculating the time as a function of the sample at
each step.

Signed-off-by: Earle F. Philhower, III <earlephilhower@yahoo.com>

### DIFF
--- a/src/output/csv.c
+++ b/src/output/csv.c
@@ -93,7 +93,8 @@ struct context {
 	uint32_t num_samples;
 	uint32_t channel_count, logic_channel_count;
 	uint32_t channels_seen;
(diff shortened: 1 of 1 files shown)


gsigh:
Merged in 01ac3eedfa23, and amended in 4feb6ec9a2c8. Thank you.

PR #64, opened 2020-05-22
Make VCD files generated by Icarus Verilog and Verilator more parseable

Niautanor:
Specifically this means:

- ignoring channels with the same identifier
  - This happens when the same channel appears in different scopes.
  - The current behavior is that the channel is listed twice but that all duplicates will stay at their initial value
- ignoring vector value changes but still parsing lines after them
  - The VCD input module currently throws away an entire chunk of data if that happens which also breaks the parsing of scalar value changes in that same chunk

### DIFF
--- a/src/input/vcd.c
+++ b/src/input/vcd.c
@@ -529,14 +529,18 @@ static void parse_contents(const struct sr_input *in, char *data)
 		} else if (strchr("bB", tokens[i][0]) != NULL) {
 			bit = (tokens[i][1] == '1');
 
+			if (!tokens[++i]) {
+				sr_dbg("Identifier missing!");
+				break;
+			}
+
 			/*
 			 * Bail out if a) char after 'b' is NUL, or b) there is
-			 * a second character after 'b', or c) there is no
-			 * identifier.
+			 * a second character after 'b'
 			 */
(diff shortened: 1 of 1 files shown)


Niautanor:
Oh, this also resolves the main complaint of bug #776.

abraxa:
Hi, thanks for your changes! Unfortunately, you didn't discuss these changes with us beforehand and so you weren't aware that there's a complete rework of the VCD input module going on. This is now merged.

What I would suggest is
a) you check whether the current VCD input module can work with the files that gave you problems
b) provide some small example files that we can perform regression tests against

If there are still issues with the files you have, feel free to update this PR against current git HEAD, as the VCD input module is in a more stable state now. Thank you!

gsigh:
The changes that you suggested in this PR are not desirable in my POV, ignoring input data doesn't look like an improvement to me. The sigrok.org code base properly addressed the issues by adding support for missing features instead of ignoring unsupported data. I'm closing this request since it has become obsolete. Feel free to report if you spot any remaining issue (and make sure to provide enough information to reproduce the issue). User feedback so far welcomed the recent feature addition.

Niautanor:
I opened this back when the current state of vcd input was still completely broken. I wanted to close this merge request once the real fix was merged but apparently missed the notification for the commend there. I am very sorry about that.

PR #65, opened 2020-06-06
serial: flush() after open() in the serial core.

wsakernel:
Quite some drivers flush the serial port after opening it. And quite
some don't although they should. Factor this out, so serial_open() will
always flush the port. The removal in the drivers was done with this
small coccinelle script:

@@
struct sr_serial_dev_inst *serial;
@@

        serial_open(serial, ...)
        ... when != serial
-       serial_flush(serial);

and then the results and the unmatched findings of serial_flush() were
audited.

Signed-off-by: Wolfram Sang <wsa@kernel.org>


gsigh:
Merged in cb828f1b3e00, thank you!

PR #67, opened 2020-06-06
openbench-logic-sniffer: improve error messages when scanning ID

wsakernel:
- always say 'ID' when the ID command failed
- print hexdump of a faulty ID because on a stalled device we may get
  0x00 bytes which would terminate the string early.

Introduce a new 'num_read' variable to avoid that in a future change
'ret' might get reused and destroy the parameter needed for hexdump.

Signed-off-by: Wolfram Sang <wsa@kernel.org>


uwehermann:
Looks good, but please drop the "ret" variable, that's unused now and causes a compiler warning.

ninjadrm:
Oops, sorry, seems I overlooked this. I updated the branch now.

uwehermann:
Merged in 35be304ba63a8d65904de2fdbcd5d848c85d7671, thanks!

PR #68, opened 2020-06-07
kingst-la2016: improved bitstream upload and fix for v3.4.2

fastflo:
of vendor's fpga bitstream.
this is now tested to work with version 3.4.0 and 3.4.2.

i realized that the data of one control-out request before the fpga-bitstream-upload is actually the bitstream-length (vendor's software uses the NON-zero-padded length here).
the version 3.4.2 bistream length is longer than version 3.4.0, thats what caused the upload to fail.

there is still another control-out request after bitstream upload that seems to depend on bitstream contents -- currently i don't understand how. i left a switch between the two currently known bitstream lengths. (so that v3.4.0 and v3.4.2 is supported)

PR https://github.com/sigrokproject/sigrok-util/pull/2 is also related to this bug. it doesn't seem to be strictly needed (it also works when zero-padded-size is specified), but i suggest to still merge it to have the same behaviour as the vendors software.

[fixes #1559]

### DIFF
--- a/src/hardware/kingst-la2016/protocol.c
+++ b/src/hardware/kingst-la2016/protocol.c
@@ -28,6 +28,7 @@
 #include <stdio.h>
 #include <errno.h>
 #include <math.h>
+#include <inttypes.h>
 #include <libsigrok/libsigrok.h>
 #include "libsigrok-internal.h"
 #include "protocol.h"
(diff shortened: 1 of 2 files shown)


uwehermann:
Merged in 3f48ab0282ea6f0b108cb441a9cc5fb76ad31e58 with two small whitespace fixes, thanks!

"if(" -> "if ("

PR #70, opened 2020-06-10
scpi-dmm: Add GW-Instek 8200A series bench meter support

tjko:
This patch adds support to GW-Instek GDM-8251A and GDM-8255A bench multimeters.

I was initially going to go with standalone driver, but noticed existing scpi-dmm driver was already implemented with excellent framework for meters using SCPI commands.

However, GW-Instek seems to interpret SCPI/IEEE488.2 differently from some other vendors like HP/Agilent/Keysight, so small change is included to "scpi" code to relax the SCPI protocol implementation (support NL+CR in addition to NL as terminator for response message) and parsing response to *IDN? command when device omits serial number... 

After studying existing code and specs, I believe the change to scpi_serial_read_data() does not break any existing SCPI device support: 236583fc190527138050caa8804b8b82fa81486a

Since currently "scpi" code does not work with any device that behaves like GW-Instek devices that terminate responses with NL+CR (instead of just plain NL that HP/Agilent devices send).  Since if any device sends anything but NL/LF character at the end of the response message, it doesn't work. Symptom being this:

```
sr: scpi: Timed out waiting for SCPI response.
```

Another change to "scpi" was to support devices that omit serial number in response to *IDN? commmand: 07aec699cfeaf509a2d99e059411572afcefc41d
(this won't have any effect on devices that respond fully SCPI compliant way to *IDN? command)

Do you feel this is good/proper way to modify the "scpi" code to allow support of vendors that interpret the spec slightly differently?  Or would it be perhaps better to make these changes conditional, so driver could toggle flag (in scpi driver context) to enable more relaxed/flexible protocol interpretation (if desired)?   Then drivers like scpi-dmm/scpi-pps could first scan for device using "strict" SCPI implementation, and if no devices found could try again with the relaxed SCPI implementation enabled?    

Or would you recommend just going the route of standalone driver for these GW-Instek devices? (I'm also looking adding support to GPD-3303S power supply that would benefit from this change, allowing it to be supported by scpi-pps driver instead of needing standalone driver as well...)


tjko:
Replaced with: https://github.com/sigrokproject/libsigrok/pull/78

PR #71, opened 2020-06-11
rigol-ds: Support for reading segmented data and some bug fixes

Cat-Ion:
Bug fixes:
- Reading was being stopped after the first frame ("End acquisition when data for all channels is acquired."). I removed that and now check the limit later.
- Sample rate can now be determined for all data sources, not only Live. It is read and stored at the start of data acquisition so it only has to be read once.
- Data transfers were missing "WAV:RES" and "WAV:BEG" calls, so sometimes the wrong data would be returned

Improvements:
- Support for reading segmented data for v3, v4, v5 protocols. Version 4 was tested, the other two should work but I don't have a matching scope to check.
- Some commands are now only called for the first frame, greatly reducing overhead for smaller transfers and many frames. For my test case of 5 frames, this halves the time needed to read 5x 14 kSa from 18 to 9 seconds.

### DIFF
--- a/src/hardware/rigol-ds/api.c
+++ b/src/hardware/rigol-ds/api.c
@@ -526,7 +526,6 @@ static int config_get(uint32_t key, GVariant **data,
 	struct dev_context *devc;
 	struct sr_channel *ch;
 	const char *tmp_str;
-	uint64_t samplerate;
 	int analog_channel = -1;
 	float smallest_diff = INFINITY;
 	int idx = -1;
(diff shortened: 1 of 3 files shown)


Cat-Ion:
Reading all segments and comparing them to the result from showing a segment and getting Live data doesn't match up, so can't merge this yet...

Cat-Ion:
Here's one such broken segment:

image

Cat-Ion:
Fixed :)

Cat-Ion:
I've tested this with my DS4000 for a bit and haven't seen any further bugs. If somebody could poke it with a DS1000Z or MSO5000 that'd be great, but otherwise this seems to be about done on my side.

uwehermann:
Merged in 418c99248cd9fe7913c3c337e831b3c601a9227a and the commits after that one, thanks a lot!

PR #74, opened 2020-06-20
fx2lafw: Add support for frames

Cat-Ion:
When using a number of frames that is not 1, the driver will read
samples up to its limit and then wait for another trigger. This will be
repeated until the configured number of frames has been finished.


abraxa:
Merged in bfc34b9ab0302fa0ba98eef3ce1b35c76d378589, thanks!

PR #75, opened 2020-06-25
scpi-pps: Add Siglent SPD3303 series PSU support

tjko:
This adds support to Siglent SPD3303C / SPD3303X / SPD3303X-E programmable DC power supplies.
Functionality was tested both via LAN (tcp-raw) and USB (USBTMC) using sigrok-cli and SmuView.

Siglent SCPI implementation is "problematic" so I implemented "quirks" in SCPI framework to allow supporting
instruments (vendors) that can't seem to follow SCPI/IEEE488.2 specifications...
This way SCPI behavior doesn't need to change for devices that comply with SCPI, but framework still can be used
with devices that require "tweaks": ca3de968497e6b0e890af74cdd26ca4d43c1178a

Several Siglent devices don't really have good USBTMC implementation, requiring delay after sending command, otherwise
unit won't respond (causing timeout). This change adds "quirk" for this behavior and enables it automatically for
affected devices (based on USB device IDs): 1a7cd9c0d784efc660dc6471e1e1ba4bcbdba9f4


tjko:
Replaced with: https://github.com/sigrokproject/libsigrok/pull/77

PR #76, opened 2020-06-27
Allow retrieving frame limits

Cat-Ion:
These patches enable querying previously set frame limits from devices that only used to store them without allowing reading them back out again. This is so we don't need to add special handling in pulseview for drivers that have this write-only

serial-lcr is curious in that it already had the `SR_CONF_LIMIT_FRAMES` case in its `config_get()`, but had not indicated this in the device options.


abraxa:
Merged in 1b8a37c688ace35683bc05b0618cb2ba353ad7e8, thanks!

PR #80, opened 2020-07-12
rigol-ds: Handle DSO1000B's CHAN#:PROB? responses

Cat-Ion:
Reported by thalesmaoa, should be tested by them before this is merged.


abraxa:
Merged in 34ce4d8258e61cb98446e027b2af0d6b522388fc, thanks!

PR #82, opened 2020-07-17
Fix for channel coupling not updating on changing, hantek-6xxx. 

wakass:
Presumed regression for bug #836


abraxa:
Merged as f5c618e55fe6e1b1bd43b441e688944c78b889b5, thank you! :)

PR #83, opened 2020-07-27
tplink-hs: Add initial support for the TP-Link HS110

fgervais:
Add support for the Kasa Smart Plug with Energy Monitoring by TP-Link.

This is basically a merge between beaglelogic and rdtech-tc.

It's my first driver in libsigrok so I'm expecting quite a few modifications might be required.


fgervais:
I did my tests using:

```
smuview -D --driver=tplink-hs:conn=tcp_raw/192.168.2.103/9999
```

src/hardware/tplink-hs/protocol.c
```diff
@@ -0,0 +1,491 @@
...
+	if (!(sdi = cb_data))
+		return TRUE;
+
+	if (!(devc = sdi->priv))
+		return TRUE;
+
+	if (revents == G_IO_IN) {
+		sr_info("In callback G_IO_IN");
+		recv_poll_data(sdi);
+		tplink_hs_tcp_close(devc);
```
fgervais, line 466: I'm pretty sure closing the socket here and keeping the session source pollfd is illegal and only works because my socket happens to get the same fd next time I connect, I'd like feedback on this.

I'm doing this because unfortunately the hs110 requires a new tcp connection for each requests and if I stay connected, the G_IO_IN event will continuously trigger. I also want to keep my event source so this callback will get called on timeout and so eventually I can poll the device again.

gsigh:
Yes if you cannot reliably keep the socket around, then you cannot use its handle for the glib mainloop. Can you create something that is not connected to the device, but remains available across the acquisition period? Like a timerfd, or some "standin" like a pipe or something, just for registration purposes. A timer may even support your period connecting to the device and getting samples (just guessing). Can't tell how portable a timer fd is though.

The sigrok-util repo has a new-driver script. Can you use that to create the first commit, please? This will tell the skeleton from the flesh, and is helpful for review and during maintenance. Also increases consistency across the drivers. Thank you!

Haven't checked details yet of the current implementation.

fgervais:
Ok skeleton done, I will now look at using a timerfd as proposed.

src/hardware/tplink-hs/protocol.c
fgervais: I switched to a combination of timer along with a temporary registration of the socket to get the reply. Now it seems legit to me.

fgervais:
The issue with the socket has been handled by the added timer source from the last commit.

gsigh:
Had a look, got some unassorted feedback: sw_limits should have config_get(), too. Socket fd probably needs to get invalidated after close(2) calls. Is shutdown(2) useful before close(2)? There is custom code for TCP communication, ideally there would be common code to not duplicate that logic too many times. Prefer common text to number conversion, the project has learned some locale(7) lessons. Prefer poll(2) over select(2), a library cannot know how many sockets an application has opened. Check error paths for balanced resources, probe may leak a connection, haven't checked others in detail. Check string literals for typos. Is the tplink-hs.h header required or can you fold it into protocol.h? Style nits: no need for deep indentation, one or two TAB are sufficient to reflect structure. Move assignments out of declaration blocks, especially the non-trivial ones. Too many empty lines may "break" code groups which actually belong together.

fgervais:
Thank you I'll address those issues

fgervais:
@gsigh I did a couple changes in the direction of your review, let me know what you think.

You might want to use the review function of github by clicking on the '+' sign right next to the line number and add you comment there while you browse my changes. This will also give me a bit more context.

gsigh:
Sorry, but no. Would rather use less of it than more, the github UI has not been a pleasant experience so far. And I'd rather not point to a single location for more general feedback that applies to the whole set of changes. This could be misleading at times. Granted that's just me, others need not agree with what I said above. Let those enjoy who do, just don't assume that all people feel in identical ways.

About the socket invalidation: I'd assume the extra boolean is not necessary. There is the integer fd that you can just void upon error or after close (assign a negative value). This approach would be more robust and less error prone. The elmination of select(2) by switching to poll(2) should be as obvious for the select() call's limitations.

gsigh:
Any progress with the source code adjustment?

fgervais:
I couldn't make much progress on this project. I'll see if I can squeeze to time on this soon.

abraxa:
No feedback, closing. Feel free to reopen if you're still interested in getting this merged, @fgervais .

PR #51, opened 2020-04-01
OLS Improvements

v1ne:
This PR adds the missing support for Advanced Triggers, as far as the sigrok trigger API can support it. Also, I took the liberty to iron out some glitches in the OLS driver.
What do you think? Is something missing? Should it be in a different shape?

PR #84, opened 2020-08-05
OLS cleanups

v1ne:
These clean-ups are taken from #51 and are the base to further refactor the driver and add support for advanced features of the DemonCore.


wsakernel:
Technically, it would be more Reviewed-by. But I don't mind much which one... Thanks!

gsigh:
Merged in 6f9234e6f333, thank you. Have added wsa's Reviewed-By. Had to adjust the commit messages. Would be nice if your future commits would stick with the project's habit of prefixing the subject with the driver name. Makes it easier to navigate later during research.

v1ne:
@gsigh: Sure, will do! Thanks for telling me.

PR #85, opened 2020-08-22
device support for R&S HAMEG HMP4040 PPS without per-channel-updates

fastflo:
this PR removes the `rs_hmp_update_status()` function from my original PR #57.

the original PR wanted to introduce a new `SR_CONF_CHANNEL_GROUP` meta package to be able to communicate down the session bus to which channel following status-updates belong to.

i find this to be a servere loss of functionality to not be able to observe CC/CV/... state changes per channel.
but i am not aware of other means to keep this information within libsigrok. compare also comment https://github.com/sigrokproject/libsigrok/pull/57#issuecomment-612838774


fastflo:
i will try to reproduce your results with sigrok-cli.
targets and measurements are separated, yes:
```
	{ SCPI_CMD_GET_MEAS_VOLTAGE, "MEAS:VOLT?" },
	{ SCPI_CMD_GET_MEAS_CURRENT, "MEAS:CURR?" },
	{ SCPI_CMD_GET_VOLTAGE_TARGET, "VOLT?" },
...	{ SCPI_CMD_GET_CURRENT_LIMIT, "CURR?" },
```
the stream should only contain measured values. i will come back after i made tests...

fastflo:
sorry, i used your branch (0805b364de859c3dca4e1ff2357ab3fe1aede369) with sigrok-cli master HEAD as of now (0e2f95c489ba93227b20747ea0ed246b51e5bf9e).

tested with my R&S HAMEG HMP4040. 

i called sigrok-cli like this: `sigrok-cli -d scpi-pps:conn=/dev/ttyUSB2 --continuous`
all output looks as i would expect:
```
with "main"-OUTPUT switch off (same with "main"-OUTPUT switch on and all
channels off):
V1: 0.0 mV DC
I1: 0.0 mA DC
V2: 0.0...

CH1 on, but "main"-OUTPUT switch off: same as above.
"main"-OUTPUT switch off, but per-channel CH1-switch on: same as above.

CH1 on & main on, CH1 at target 2V, nothing connected:
V1: 2.0000 V DC
I1: 0.2 mA DC
V2: 0.0...

with 1kOhm on output:
V1: 2.0000 V DC
I1: 2.2 mA DC
V2: 0.0...

with current limit at 10mA and shortcut on output:
V1: 3.0 mV DC
I1: 10.1 mA DC
V2: 0.0...

with target voltage at 15V and current limit at 10mA and 1kOhm on output:
V1: 10.2100 V DC
I1: 10.1 mA DC
V2: 0.0...
```

also `sigrok-cli -d scpi-pps:conn=/dev/ttyUSB2 -C V1,I1 --continuous` looks as expected:
```
V1: 9.8410 V DC
I1: 9.7 mA DC
...
```
setting/getting "config" also seems to work as expected:
```
sigrok-cli -d scpi-pps:conn=/dev/ttyUSB2 -g 1 --get voltage_target
15.0
sigrok-cli -d scpi-pps:conn=/dev/ttyUSB2 -g 1 -c voltage_target=10 --set
sigrok-cli -d scpi-pps:conn=/dev/ttyUSB2 -g 1 --get voltage_target
10.0
sigrok-cli -d scpi-pps:conn=/dev/ttyUSB2 -g 1 --get voltage
0.0

with output enabled:
sigrok-cli -d scpi-pps:conn=/dev/ttyUSB2 -g 1 --get voltage
9.9979999999999993

with output enabled & 1kOhm:
sigrok-cli -d scpi-pps:conn=/dev/ttyUSB2 -g 1 --get voltage --get current
10.0
0.0099000000000000008

sigrok-cli -d scpi-pps:conn=/dev/ttyUSB2 -g 1 -c voltage_target=15 --set
sigrok-cli -d scpi-pps:conn=/dev/ttyUSB2 -g 1 --get voltage --get current
9.9809999999999999
0.0097999999999999997

sigrok-cli -d scpi-pps:conn=/dev/ttyUSB2 -g 1 --get voltage_target --get current_limit
15.0
0.01
```

`sigrok-cli -d scpi-pps:conn=/dev/ttyUSB2 --show` output:
```
Driver functions:
    Power supply
Scan options:
    conn
    serialcomm
scpi-pps - HAMEG HMP4040 HW50020001/SW2.62 [S/N: 0181_____] with 8 channels: V1 I1 V2 I2 V3 I3 V4 I4
Channel groups:
    1: channels V1 I1
    2: channels V2 I2
    3: channels V3 I3
    4: channels V4 I4
Supported configuration options across all channel groups:
    continuous: on, off
    limit_samples: 0 (current)
    limit_time: 0 (current)
```

can you provide more details exactly which config options/parameters you are changing and do not behave as expected?

in a few weeks i should be able to try to make photos of my device (its currently buried in a test-setup).

gsigh:
Must have been staring at 'sigrok-cli -d scpi-pps --show' and '... --show -g 1' for too long, could not see the currently configured voltage and current settings. It seems that the parameters' ranges are reflected instead of their current value (could be an application side issue in the sigrok-cli code base, haven't checked details).

When I tried other commands, the libsigrok driver worked as expected: 'sigrok-cli -d scpi-pps -g 1 --get voltage_target --get current_limit', and '--samples 20' with optional '-C V1,I1', etc.

One minor issue is that the device remains in remote mode even after use, when the application no longer is connected. This can get addressed in a later commit when the cause got determined.

fastflo:
> One minor issue is that the device remains in remote mode even after use, when the application no longer is connected. This can get addressed in a later commit when the cause got determined

yes, as soon as it receives an scpi command it switches into remote. but there is a command to explicitely switch it back to local. this could be sent do the device after the acquisition as a last command. but i am not yet sure where to place this additional command. is there a per-device deinit-hook which might also be called from a SIGINT handler? if acquisition is limited by number of samples or similar, it should be easy enough to do it after reading the last sample...

gsigh:
Merged in bd5f0a143eb1 with some style fixes on top, thank you. Have also added "all other" HMP4000 devices since their construction is so modular and all of them should work.

Have added support for remote and local operation. Was not comfortable with "mixed" so I dropped it for now. It's easy enough to adjust locally for those who want it.

PR #88, opened 2020-09-09
Fix for Windows cross-compile error due to Windows headers defining D…

x0rloser:
…IFFERENCE and RELATIVE


uwehermann:
Hi, thanks for your PR!

I've fixed this in https://github.com/sigrokproject/pulseview/commit/4788af1e3e48461cc60d1fc035e5de64e914c03d using a slightly different method, avoiding the namespace pollution from windows.h altogether. This could potentially also help with other random conflicts in the future.

PR #89, opened 2020-09-15
tests: strutil, use ck_assert

jirislaby:
We pass no message, so use `ck_assert` instead of `ck_assert_msg`. This
results in an error with check 0.15:
```
tests/strutil.c:157:2: error: too few arguments to function '_ck_assert_failed'
  157 |  ck_assert_msg(saved_locale != NULL);
```


gsigh:
Merged in 505a55f7899e, thank you.

PR #90, opened 2020-09-17
Added support for importing NCI GoLogic and GoLogic XL project files.

x0rloser:


gsigh:
Do you have an interest in getting this feature included into mainline sigrok? Check the doc files in the source tree's root directory to improve the odds.

x0rloser:
Yes I would very much like to get this into mainline. Can you be more specific about what doc files I need to read or what changes are needed?

Around the time I submitted this pull request I was in the IRC channel for a few weeks before and after submitting. I brought it to the attention of the ops there are was advised told to add the sigrok license headers but nothing more.

biot:
Sorry to see you never got this properly reviewed. Are you still interested in getting this merged? If so, I can help with getting this reviewed and merged. A few preliminary comments:

- The patches need squashing, and rebasing to current master.
- The code is in really rough shape. Coding style doesn't match sigrok's, check https://github.com/sigrokproject/libsigrok/blob/master/HACKING for more info.
- This compiles with a TON of warnings. Did you not see those somehow? libsigrok should compile with no compiler warnings at all.

Also, the code has this note:
```
 * LIMITATIONS
 * - Sigrok code only supports channels but gologic uses "groups" which can
 *   have 0, 1, or multiple channels. When importing, any group with 0 or multiple
 *   channels will be ignored. So only groups with exactly 1 channel will get imported.
 * - Gologic supports up to 72 channels but sigrok only supports 64 max. So if all
     channels are used in a project, sigrok will not be able to handle more than 64 of them.
```

Literally none of that is true. sigrok supports an unlimited number of channels, and channel groups with an unlimited number of channels.

PR #93, opened 2020-09-23
rigol-ds: Add 200ns timebase value

jr-oss:
The 200ns timebase value was missing for rigol-ds devices.


jr-oss:
The one-liner has been integrated in PR#95

PR #94, opened 2020-09-23
scpi: accept numbers like 4.0000E3 as integer value

jr-oss:
MSO5000 returns memory depth value in that format, e.g.
sr: [04:21.491949] scpi_vxi: Successfully sent SCPI command: 'ACQ:MDEP?'.
sr: [04:21.501463] scpi: Got response: '4.0000E+03', length 10.

I am not sure if it is acceptable to change the implementation of "sr_scpi_get_int" to get it working with Rigol MSO5000 or if it would be better to extend the scpi-API.

### DIFF
--- a/src/scpi/scpi.c
+++ b/src/scpi/scpi.c
@@ -724,6 +724,7 @@ SR_PRIV int sr_scpi_get_int(struct sr_scpi_dev_inst *scpi,
 			    const char *command, int *scpi_response)
 {
 	int ret;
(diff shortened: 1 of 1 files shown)


gsigh:
Merged in 2dddd5bd5edc, thank you. Have squashed minor tweaks on top. The strict check for the value being an integer made it acceptable, and the SCPI doc explicitly allows devices to respond that way. So it's a clear improvement in what was incomplete before.

PR #97, opened 2020-10-25
add support for ADALM-2000

trabucayre:
[Analog Device ADALM2000](https://wiki.analog.com/university/tools/m2k) is a data acquisition module:
- 2 channel 12bits 100MS/s oscilloscope 
- 2 channel 12bits 100MS/s arbitrary function generator
- 16 channel 100MS/s logic analyzer
- 16 channel 100MS/s pattern generator (uses same pins as logic analyzer)

This PR add a basic support for this device:
- logic analyzer support tested at 5MS @ 100MS/s 
- pre trigger delay
- basic trigger (only one channel)
- one-shot mode only

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -213,6 +213,14 @@ src_libdrivers_tail_la_SOURCES = src/driver_list_stop.c
 
 src_libdrivers_la_SOURCES = src/drivers.c
 
+if HW_ADALM_2000
+src_libdrivers_la_SOURCES += \
+	src/hardware/adalm-2000/m2k_wrapper.h \
+	src/hardware/adalm-2000/m2k_wrapper.cpp \
+	src/hardware/adalm-2000/protocol.h \
+	src/hardware/adalm-2000/protocol.c \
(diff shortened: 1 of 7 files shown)


teoperisanu:
This is an awkward situation. We have planned for a long time to create a driver for adalm2000 and we started the development in march. The version Robin referred to, is almost done, with just some minor fixes needed. We never published the driver because we waited for the official firmware (https://github.com/analogdevicesinc/m2k-fw/releases/tag/v0.26) that includes some of the functionalities we wanted in the driver.
I will mention some important functionalities present in the driver. Both the digital and analogical inputs are integrated and these are synchronized so that when both are running a mixed signal acquisition is performed. The triggers for both components are configurable and many others attributes, such as vertical offset, the analogical range, the trigger delay and many others.
I don't know what is the further procedures but I wanted to mention these things.

trabucayre:
I not found this repository before, nor references in wiki.analog.com. This why I've written this support...
I suppose my only option is to close this PR...

PR #100, opened 2020-11-05
Driver for etommens etm xxxxp and clones

SpeedProg:
my first time writing a driver for sigrok so probably a good idea to have a good lookover.
I use g_malloc0 twice to allocated memory for the device context struct and sr_dev_inst.
I would have assume I need to deallocate these, but since no other driver seems to do that, I just let it be.
It only supports the RS310P, HM3010P and eTM-3010P since I don't know the right model and class ids for other ones.


SpeedProg:
updated formatting after some one on irc pointed out some inconsistencies and that I should indent broken up lines with 2 tabs instead of the kernel coding styles "substantially to the right"

tomtastic:
I believe the HM305P reports the following for it's model and class :
```
Model (0x03): 0x0BC2
Class (0x04): 0x4B50
```

SpeedProg:
Added the 30V 5A model data

thehans:
Hello, I just purchased a Hanmatek HM310**T** which is I think the latest iteration of design for the 30V 10A version of this PSU.  

It looks like it uses the same communications interface as a HM310P, etc just with a different front faceplate / display 
(mainly there's an additional 8 segment display that counts seconds while output is enabled)
It has a NUCO29LAN chip inside which is same as 3XXP according to sigrok wiki, and the software/docs that came with it talk about the modbus registers etc.  I'm not sure yet if it has any *additional* parameters compared to 310P (e.g. maybe the seconds counter).

The power board is a little different, and there are some teardown pics here if anyone is curious.
https://www.eevblog.com/forum/testgear/power-supply-ripe-for-the-picking/msg3557624/#msg3557624

I've run the `py_test_interface` patch from @tomtastic and this is the same as what an HM310P would return as far as I know:
```
Model (0x03): 3010
Class (0x04): 0x4b50
12.44 Volts
0.0 Amps
0.0 Watts
```
The volts/amps/watts from that script also match exactly with the segmented display readings which is a good sign.

I have not attempted to build this PR yet(i plan to soon), but I think it would be great if the changes could get reviewed and merged.

If there's anything specific that I can do to help test this model, please let me know.

thehans:
In case someone wants to see a product page for reference, mine was purchased from amazon and it can be found by searching "HM310T" there.  I would provide a link but don't want to appear as advertising / endorsing any particular distributor.

There are probably other avenues of purchase, although additional info on this model did seem rather scarce when I looked.  For example I did not see it on eTommens site which is allegedly the original manufacturer of the other models.

thehans:
@SpeedProg I had some discussion in IRC (#sigrok on freenode) regarding whether this PR could be merged.

I was told that it is not acceptable as is.

Reasons given to me were:
1) Parts of code appeared too similar to (allegedly copied from) some other sources without sufficient attribution.
2) There is now a modbus driver in master.  It sounds like it didn't exist at time of original PR, but would need to integrate with that.  Quote: "the PR fell behind master, should fold into the common modbus psu driver"
3) Something about "layering violations" (maybe would be solved by #2 anyways?)

I figured I would relay this information, because the individual I spoke with prefers not to use github.  So it seems you are not likely to get an official code review here, at least not from them.  (Note: this repo is just a mirror and the main repo is hosted elsewhere)
Also just to be clear: I am NOT a sigrok maintainer, just an interested party :smile:.

I may try looking into this common modbus psu driver myself and see what I can do though, as I would really like to see these devices supported.

SpeedProg:
1. They are right, I was sure I added copyright reference for them. But apparently I didn't, which was very very very bad on my part. They are right, I was sure I added copyright reference for them. But apparently I didn't, which was very very very bad on my part. Updated the PR with what I remember to be correct copy right attribution even if this code is not gonna get merged like this anyway.
2. There was some SR modbus things when I started writing the code, I will further look into it when I get the time.
3. It means I am not properly separating what I am doing in protocol.c (data acquisition) from api.c, I am not entirely sure how they want it separated through.

knarfS:
Please have a look at the recently updated `rdtech-dps` driver. It has hugely improved, compared to the old one (on which this driver is based). It should be fairly easy now to add new modbus devices to the existing `rdtech-dps` driver.

SpeedProg:
I took a look today and if you could answer them, I would have a few questions about how to integrate it.
1. The device does not support multi-register update (0x10), so I needed, for this code already, to implement 0x06  (single register update) for it, should I move that in a patch to the sigrok modbus module instead?
2. There is the [rdtech_dps_model](https://github.com/sigrokproject/libsigrok/blob/c899e691f3380f25eebeac8afe630e5c657e7c61/src/hardware/rdtech-dps/protocol.h#L41) should I expand that if needed or just create a different model (same with the device context actually), which if creating a different one would probably make me unable to use most of the existing functions. But their names don't really imply that you should use them or anything then rdtech hardware.

thehans:
@SpeedProg I know some of the documentation that comes with the PSU claims that only Function codes 3 & 6 are supported, but I am fairly certain that it actually *does* support function code 16 for multiple register write.  

The reason I think this is because I had stumbled across the [PyModbus REPL](https://pymodbus.readthedocs.io/en/latest/readme.html#pymodbus-repl-read-evaluate-print-loop) where I was able to get it to connect and try out various commands.  When I tried [`client.write_registers`](https://pymodbus.readthedocs.io/en/latest/source/library/pymodbus.client.html#pymodbus.client.common.ModbusClientMixin.write_registers), it worked and responded just fine.  I can only assume that corresponds with function code 16, although I didn't see the raw message that pymodbus sent based on the command.

Another reason I had suspected that might work, is because I saw a reference to function code 16 in the included software/XML files.  This was actually one of the things I found in comparing a diff of the software distribution that came with my unit, and a HM305P download I found online.  So, *maybe* it only works on this version(HM310**T**), but I have a feeling it should work on the others since I couldn't find any other single difference in terms of communication protocol.

By the way, I did a whole lot of snooping around the protocol on my HM310T, mostly via markrages' `py_test_interface/hm305.py` script and making many additions and modifications of my own ... although I hadn't gotten around to implementing Function code 16 just yet.
I did at least optimize the way it reads multiple registers, and ended up making a function to scan the whole register address space for non-zero values.  I found quite a few "undocumented" ones (well let's be frank, they're ALL undocumented).  Although nothing particularly interesting as they seemed to be either mostly duplicated data from other known registers, or just more readonly constants.

Anyways I did my best to document as much as I could understand from my adventures, and put it in the code and comments of my fork here, in case its any use to you:
https://github.com/thehans/py_test_interface/blob/master/hm305.py

I had also posted about my "reverse engineering" efforts in the eevblog thread [here](https://www.eevblog.com/forum/testgear/power-supply-ripe-for-the-picking/msg3560110/#msg3560110), if you care for a bit more narrative.

SpeedProg:
I am fairly certain I tried to write to mine with 0x10 back then and it was not working.
I got most of my info from the windows software that was delivered with the station (C# is pretty easy to decompile) and the xml configuration file that were there with it, when afterwards I like you read some other registers and tried to figure out what they are used for. What I found noteworthy was documented on the sigrok wiki.

SpeedProg:
You are right it seems to work with the modbus 0x10 code, so I guess I either didn't try or my code was bugged back then.

knarfS:
>     1. The device does not support multi-register update (0x10), so I needed, for this code already, to implement 0x06  (single register update) for it, should I move that in a patch to the sigrok modbus module instead?

If you can't use multi register writes as @thehans suggested, then go ahead and add the necessary function to modbus.c (f.e. something like `sr_modbus_write_single_register(...)`).

>     2. There is the [rdtech_dps_model](https://github.com/sigrokproject/libsigrok/blob/c899e691f3380f25eebeac8afe630e5c657e7c61/src/hardware/rdtech-dps/protocol.h#L41) should I expand that if needed or just create a different model (same with the device context actually), which if creating a different one would probably make me unable to use most of the existing functions. But their names don't really imply that you should use them or anything then rdtech hardware.

I think it would make sense to add generic stuff to the `rdtech_dps_model` struct.
The function-, struct- and enum-prefixes (`rdtech_dps_`) are related to the original driver name (and now just the source directory). Renaming the whole thing is maybe reasonable in a next step.

SpeedProg:
Thank you.
Gonna do the work for the driver rewrite on https://github.com/SpeedProg/libsigrok/commits/etommesn-rdtech before it's ready to update this pullrequest (I gonna force push into it a lot so please don't base anything you do of that branch) if anyone wants to take a look at the status, but it's gonna take me probably quite a while to finish it since I don't have that much time. So if anyone is faster, feel free to do it.

SpeedProg:
I pushed the new driver implementation that is integrated into the rdtech driver

dperruso:
Hi everyone,
i'm chiming in because i would love to see this PR merged. Why isn't it already?

biot:
Sorry it took so long to get this reviewed. I'd like to help review and get this merged.

However the rdtech-dps driver has seen quite a few changes since you wrote this, and your patch no longer cleanly applies. Could you rebase on top of master? Thanks.

SpeedProg:
I will do that. It is gonna take a week or two through, because I've got no time this weekend.

ensc:
> I will do that. It is gonna take a week or two through, because I've got no time this weekend.

fwiw, https://github.com/ensc/libsigrok/commit/aa6f43660540a396257fd99061a5eec4a84458a6 should be an updated patch

SpeedProg:
I updated the patch now, but I could not mange to test it. Since if I build sigrok-cli against the current HEAD of master, sigrok-cli claims it has no hardware drivers. Where I have my patch applied or not.
Not sure if that is a me problem or if they are just incompatible right now.

biot:
> I updated the patch now, but I could not mange to test it. Since if I build sigrok-cli against the current HEAD of master, sigrok-cli claims it has no hardware drivers. Whether I have my patch applied or not. Not sure if that is a me problem or if they are just incompatible right now.

It's not you... a bug snuck in which caused this. It's being discussed in #252 right now. Will be ok in a day or so.

biot:
The bug is fixed now, rebasing on top of current master should fix things.

tomtastic:
I must be missing a step, can't get it to recognise my Hanmatek HM305P ?

SpeedProg:
> I must be missing a step, can't get it to recognise my Hanmatek HM305P ?
> [MAC] ~/src/libsigrok on  driver/etommens-etm-xxxxp took 2s ❯ sigrok-cli -d rdtech-dps:conn=/dev/tty.usbserial-110 -l 5 --scan

You are using the wrong driver, use "etommens_etm_xxxxp" not "rdtech-dps" after compiling with this PR patched in.

tomtastic:
Heh, oops!
```
[MAC] ~/src/libsigrok on  driver/etommens-etm-xxxxp took 2s ❯ sigrok-cli -d etommens_etm_xxxxp:conn=/dev/tty.usbserial-22240 --scan
The following devices were found:
etommens_etm_xxxxp - RockSeed etM-305P/RS305P/HM305P etommens_etm_xxxxp with 3 channels: V I P
```

tomtastic:
I see digits_power for this model is correctly `3`, but the output is only `2` :

eg. Units display for power is currently `1.989W`, but the data is rounded to 2 decimal places.
```
sr: [00:01.191118] rdtech-dps: probe: ret 0, modelid 305, dclassid 19280, limit_voltage 2, limit_current 2, digits_voltage 2, digits_current 3, digits_power 3
FRAME-BEGIN
V: 4.90 V DC
I: 406 mA DC
P: 1.99 W
FRAME-END
```

tomtastic:
`--get regulation` seems fixed on `CV`, and doesn't follow the actual mode when the power supply is in `CC` mode.

SpeedProg:
Thank you, I will look into both of these.
Maybe I got something wrong about the CC/CV states.

SpeedProg:
@tomtastic 
Hey I you were right the output state 'CC' 'CV' was broken, I had gotten the wrong register.

As for your second problem with setting current you want to set current_limit not current. Current = what it currently outputs, current_limit = where it  would go into CC mode.
So the error message you got is correct, not applicable :)

PR #101, opened 2020-11-08
new SR_CONF_CHANNEL_GROUP meta package

fastflo:
many meta packages can not be assigned to a specific channel(-group). but there
are many devices like multi-channel PPS that report state information per
channel.
so instead of extending those existing meta packages with a new field to
specify the channel, i propose to add this new meta package that will only
transport the "current" channel-group-name.
all following meta packages should then be interpreted to only describe
the specified channel group.
i demonstrate this here with the HMP4040 SCPI-PPS which reports these meta
packages per channel-group:

- SR_CONF_OVER_VOLTAGE_PROTECTION_ACTIVE
- SR_CONF_OVER_TEMPERATURE_PROTECTION_ACTIVE
- SR_CONF_REGULATION
- SR_CONF_ENABLED

all those packages are only sent on change event.

i also did send a PR to demonstrate how SmuView could use this additional
information to correctly display per-channel-state info.

originally this was part of PR https://github.com/sigrokproject/libsigrok/pull/57

### DIFF
--- a/include/libsigrok/libsigrok.h
+++ b/include/libsigrok/libsigrok.h
@@ -1113,6 +1113,9 @@ enum sr_configkey {
 	/** Number of powerline cycles for ADC integration time. */
 	SR_CONF_ADC_POWERLINE_CYCLES,
 
+	/** Specify which channel group following meta packages should be assigned to. */
+	SR_CONF_CHANNEL_GROUP,
+
 	/* Update sr_key_info_config[] (hwdriver.c) upon changes! */
 
 	/*--- Acquisition modes, sample limiting ----------------------------*/
--- a/src/hardware/scpi-pps/api.c
+++ b/src/hardware/scpi-pps/api.c
@@ -101,6 +101,9 @@ static struct sr_dev_inst *probe_device(struct sr_scpi_dev_inst *scpi,
 	devc = g_malloc0(sizeof(struct dev_context));
 	devc->device = device;
 	sr_sw_limits_init(&devc->limits);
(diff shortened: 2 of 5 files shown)


fastflo:
corresponding changes in SmuView: https://github.com/knarfS/smuview/pull/15

biot:
I see what you mean wrt some of these meta keys needing to be per-channel, clearly that's what the hardware means. However:

- This changes the sigrok datafeed API in an exceptionally underhand way: it doesn't change the APi headers, or the ABI of feed packets -- but clients now suddenly need to interpret packets differently. There is no way for an outdated client that doesn't know about your new meta key to figure out anything changed. This is very, very, very bad. When you change API, you have to give clients a chance to know about it.
- You call it SR_CONF_CHANNEL_GROUP, but in fact you're dropping a channel in there, not a channel group.
- to top it off, it's the channel name, when we have these nice channel and channel group structs.

But really my first point makes this unacceptable.

Still the functionality is needed. I propose just adding a `GSList *channels` or a `struct sr_channel_group` to the `sr_datafeed_meta` packet, in any case normally NULL unless applicable. That's an APi change, but backward compatible -- since we're adding a field to the end of the struct, outdated clients will still work with the packet.

According to [semantic versioning](https://semver.org/) principles, that's a MINOR increase of the libsigrok version. I think it's worth it.

@abraxa you want to weigh in here?

abraxa:
I agree with your points, @biot. I would prefer to do this after the upcoming release though, since there's enough stuff pending to warrant a release already.

PR #102, opened 2020-11-11
appadmm-stage

Cymaphore:


biot:
Martin,

I'm sorry to see this driver never got reviewed, particularly since it supports such an enormous number of devices. Are you still interested in getting this merged? If so, I'd like to help in getting this reviewed and merged.

A few things that would help:
- Can you squash all the driver commits? Normally a new driver needs just two commits: one stub generated by the `new-driver` script in sigrok-util, and another one filling out the actual driver code. This makes it much easier to review than the ton of commits and merge entries you have now.
- Can you separate out the csv output module and other non-driver changes into separate PRs? I'll handle those as well.
- I'm concerned by the following in the wiki: "_If a specific feature, for example data logging, is not supported by a specific device, it is for compatibility reasons visible in the driver but will return an empty result._" This is really not ok, particularly since most devices support some kind of identification in their protocol. The driver should scan and identify, and then report exactly what features it supports. Lots of drivers do it this way. If a device really cannot be identified with a scan, you can use the [force_detect](https://sigrok.org/wiki/Connection_parameters#Forced_detection) parameter to let the user ID it. Clunky, but it works.

I hope you'll work with me to get this excellent driver merged.

Great work on documenting all this in the wiki, BTW.

PR #107, opened 2020-11-29
Minor bug fixes

rafl:


wsakernel (changes requested):
Patch 1 not only fixes a leak, but the code is also a lot better to read. Patch 2 is likely to be correct, too. It seems unlikely to me that in this one case the protocol says to not check the retval. So, codewise the patches are fine. Just the patch descriptions need prefixes. I.e. "beaglelogic: Fix resource leak". Use 'git log' on the changed files to learn about the proper prefixes.

abraxa:
Thanks for the review, @wsakernel !

Commits have been pushed as a99ff45359a590739c8a66458511c64d00033a5d and b41562f6ce3ba5572d65965811e0eb1f62a4b223, respectively. Thank you, @rafl !

PR #109, opened 2020-12-05
hantek-dso: fix 100MS/s

albertoscotta:
Do not downsample at 100MS/s, because this is the base frequency.
The device actually sampled at (100.0/0x1000)MS/s when 100MS/s was set.
Tested to be working on an Hantek DSO-2250 USB.


abraxa:
Merged as fe185e4990edba03faf6485dcd6a42aeb3b9a008, thank you!

PR #110, opened 2020-12-11
zketech-ebd-usb: mutliple fixes and upgrades

cuvoodoo:
- fix load toggling at acquisition stop
- messages are sent byte per byte with a pause in between since bytes were sometimes dropped
- incoming stream is parsed continuously instead of using a fixed length (the messages have a constant start/stop)
- add under-voltage threshold (under which the load would stop) as configuration
- add D+/D- voltage measurements as two new channels (and has to rename the existing to avoid confusion)


wsakernel (changes requested):
Could you kindly split this one commit into multiple ones with one commit fixing one issue? This makes reviewing much easier (and reverting in case of regressions).

cuvoodoo:
splitting the patch would require too much time at this point. I started by restructuring the protocol in order to extend it when I found and fixed the issue, which I integrated in the extension.
having the fix separate from the extension would require a complete rewrite.
I'm happy to be contacted if there is an issue with this patch.

knarfS:
Thanks for your PR!

Can you please have a look at my commit https://github.com/knarfS/libsigrok/commit/e5b57009bb5f4671eb1b92fa12d22f498c596440 where I
- fixed some style issues
- removed unnecessary code
- renamed `voltage_limit` -> `uvc_threshold` because in the sigrok context `voltage_limit` is something else :)
- fixed swapped VBUS current and VBUS voltage channels

abraxa:
Thanks for your contributions! I merged both as ab3e9c8af67b5919dd7cfb3119081bfd5858cc81 and following.

PR #111, opened 2020-12-17
fx2lafw: Make SR_CONF_LIMIT_SAMPLES above 2^32 working

jr-oss:
Sampling with "sigrok-cli -d fx2lafw -o capture.sr --config samplerate=8M --time 600000" does not stop after 10 minutes
8M*600s is 4,800,000,000 samples, which needs more than 32-bit in devc->sent_samples for counting


gsigh:
Merged in libsigrok 6065d660e04b. Thank you.

PR #119, opened 2021-02-09
python bindings: Support passing in rational voltages for config_set

rsaxvc:
This was needed to allow python to adjust SR_CONF_VDIV

### DIFF
--- a/bindings/python/sigrok/core/classes.i
+++ b/bindings/python/sigrok/core/classes.i
@@ -340,8 +340,15 @@ Glib::VariantBase python_to_variant_by_key(PyObject *input, const sigrok::Config
         return Glib::Variant<double>::create(PyFloat_AsDouble(input));
     else if (type == SR_T_INT32 && PyInt_Check(input))
         return Glib::Variant<gint32>::create(PyInt_AsLong(input));
-    else
-        throw sigrok::Error(SR_ERR_ARG);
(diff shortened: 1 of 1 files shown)


abraxa:
For the record, causes a compile-time warning similar to https://github.com/swig/swig/issues/1379

sigrok/core/classes_wrap.cpp:3963:37:   required from ‘int swig::asptr(PyObject*, Type**) [with Type = std::map<std::__cxx11::basic_string<char>, Glib::VariantBase>; PyObject = _object]’
-sigrok/core/classes_wrap.cpp:36373:176:   required from here
+sigrok/core/classes_wrap.cpp:36380:176:   required from here
 sigrok/core/classes_wrap.cpp:4046:8: warning: ‘void* memset(void*, int, size_t)’ clearing an object of type ‘struct std::pair<std::__cxx11::basic_string<char>, Glib::VariantBase>’ with no trivial copy-assignment; use assignment instead [-Wclass-memaccess]
   memset(v_def,0,sizeof(Type));
   ~~~~~~^~~~~~~~~~~~~~~~~~~~~~

Apparently, this has been fixed in swig in the meanwhile, so shouldn't be of concern for libsigrok.

abraxa:
Merged as a2916ad01779e1cf8a5e220eb6288dab9734b146, thank you! :)

PR #122, opened 2021-02-21
Enable C++17 language features to satisfy libsigc++ requirements

amotl:
Dear Uwe, Bert, Gerhard and all other contributors,

first things first: Thanks a stack for conceiving and maintaining libsigrok.

While working on https://github.com/sigrokproject/sigrok-meter/pull/1, when trying to build on macOS Catalina with recent compilers (GCC 10.2.0) and libraries (libsigc++ 3.0.6) installed through Homebrew, the process croaked with `error: 'invoke' is not a member of 'std'`.

It turned out that `libsigc++` is telling us [1]:

> Compatible compilers must support C++17, such as the `decltype(auto)` specifier (from C++14) and `std::invoke()`.

So, this brings in C++17 support by updating the `ax_cxx_compile_stdcxx.m4` macro to the most recent version from [2] and by adjusting some bits in `configure.ac`. Please bear with me when I might have done something wrong, as C/C++ is not my native habitat. However, with those changes, GCC stopped croaking on my end.

With kind regards,
Andreas.

P.S.: Others reading here also aiming at building on a recent macOS, please also note #123.

[1] https://github.com/libsigcplusplus/libsigcplusplus#compatibility
[2] http://git.savannah.gnu.org/gitweb/?p=autoconf-archive.git;a=blob_plain;f=m4/ax_cxx_compile_stdcxx.m4

PR #123, opened 2021-02-21
Add compatibility for glib>=2.62 (replace GTimeVal by GDateTime)

amotl:
Hi again,

while working on https://github.com/sigrokproject/sigrok-meter/pull/1, when building on macOS Catalina with glib installed through Homebrew, there is already glib 2.66.7.

However, `GTimeVal` has been deprecated since glib 2.62 and should be replaced by `GDateTime` [1], see also [2].

> `GTimeVal` has been deprecated since version 2.62 and should not be used in newly-written code. Use `GDateTime` or `guint64` instead.

So, this is just a humble attempt to make things work again on this end. As I told you in #122, I am not deeply involved into C/C++, but this update makes things work for us so you might want to build upon it in one of the future iterations.

With kind regards,
Andreas.

[1] https://developer.gnome.org/glib/stable/glib-Date-and-Time-Functions.html#GTimeVal
[2] https://tecnocode.co.uk/2019/08/24/gtimeval-deprecation-in-glib-2-61-2/

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -71,6 +71,7 @@ SR_PKG_VERSION_SET([SR_PACKAGE_VERSION], [AC_PACKAGE_VERSION])
 SR_LIB_VERSION_SET([SR_LIB_VERSION], [4:0:0])
 
 AM_CONDITIONAL([WIN32], [test -z "${host_os##mingw*}" || test -z "${host_os##cygwin*}"])
+AM_CONDITIONAL([MACOS], [test -z "${host_os##darwin*}"])
 
 #############################
 ##  Optional dependencies  ##
@@ -371,7 +372,9 @@ AS_IF([test "x$HAVE_CXX11" != x1],
 	[SR_APPEND([sr_cxx_missing], [', '], ['C++11'])])
 
 # The C++ bindings need glibmm.
-SR_PKG_CHECK([glibmm], [SR_PKGLIBS_CXX], [glibmm-2.4 >= 2.32.0])
+AM_COND_IF([MACOS],
+	[SR_PKG_CHECK([glibmm], [SR_PKGLIBS_CXX], [glibmm-2.68 >= 2.32.0])],
+	[SR_PKG_CHECK([glibmm], [SR_PKGLIBS_CXX], [glibmm-2.4 >= 2.32.0])])
 AS_IF([test "x$sr_have_glibmm" != xyes],
 	[SR_APPEND([sr_cxx_missing], [', '], [glibmm])])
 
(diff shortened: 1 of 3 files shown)


configure.ac
```diff
@@ -371,7 +372,9 @@ AS_IF([test "x$HAVE_CXX11" != x1],
 	[SR_APPEND([sr_cxx_missing], [', '], ['C++11'])])
 
 # The C++ bindings need glibmm.
-SR_PKG_CHECK([glibmm], [SR_PKGLIBS_CXX], [glibmm-2.4 >= 2.32.0])
+AM_COND_IF([MACOS],
+	[SR_PKG_CHECK([glibmm], [SR_PKGLIBS_CXX], [glibmm-2.68 >= 2.32.0])],
+	[SR_PKG_CHECK([glibmm], [SR_PKGLIBS_CXX], [glibmm-2.4 >= 2.32.0])])
```
amotl, line 377: On macOS/Homebrew, `glibmm` is named differently.

```sh
$ pkg-config --list-all | grep glib
gio-unix-2.0                        GIO unix specific APIs - unix specific headers for glib I/O library
gio-2.0                             GIO - glib I/O library
glib-2.0                            GLib - C Utility Library
glibmm-2.68                         glibmm - C++ wrapper for GLib
```

amotl:
> ... build upon it in one of the future iterations.

While the current implementation within this patch works for me, I am not sure if it will produce a memleak.

> GDateTime is reference counted: the reference count is increased by calling `g_date_time_ref()` and decreased by calling `g_date_time_unref()`. When the reference count drops to 0, the resources allocated by the `GDateTime` structure are released [1].

> You should release the return value by calling GLib.DateTime.unref() when you are done with it [2].

---

On the other hand, when already refactoring something related to timekeeping, using the monotonic clock might also be an option [3,4]?

[1] https://developer.gnome.org/glib/stable/glib-GDateTime.html
[2] https://lazka.github.io/pgi-docs/GLib-2.0/classes/DateTime.html#GLib.DateTime.new
[3] https://stackoverflow.com/questions/41894213/what-is-the-use-of-clock-realtime
[4] https://stackoverflow.com/questions/3523442/difference-between-clock-realtime-and-clock-monotonic

depili:
Implemented in https://github.com/sigrokproject/libsigrok/commit/e446b3c461f88a85298fe6eec294b4906909c50b and https://github.com/sigrokproject/libsigrok/commit/ee9e086f1d6c9a43a19426045d150379a40ba779

amotl:
Dear @depili,

thank you very much.

With kind regards,
Andreas.

PR #125, opened 2021-02-28
scpi-dmm: Add support for Owon XDM2041

PetteriAimonen:
Wiki page: https://sigrok.org/wiki/Owon_XDM2041

I had to add two quirks:

- Autodetection whether OPC command is available.
- Model specific infinity limit. For Owon, infinity is as low as 1e9 for all modes.


PetteriAimonen:
Merged in 2111d15

PR #126, opened 2021-02-28
Fixed PICkit2 driver and added support for PICkit3

ean365:
I did a bunch of work on the Microchip PICkit logic analyzer driver.  This driver was initially written by Gerhard Sittig for the PICkit2, but had several unresolved issues.  Much of his original work remains.  This would have been a much bigger task for me, one which I probably would not have bothered to undertake on a lazy day, had it not been for the existence of Gerhard's initial efforts.  So, thanks to Gerhard.

I fixed up the sections necessary to make the driver work properly with the PICkit2, while also adding support for the PICkit3, and all of their logic analyzer features - including all integer multiple sample rates, all sample ratios, and display of trigger position. Trigger count, triggering after N number of triggers, is also implemented (look for `trig_count`), but won't be complete until sigrok-cli and pulseview have a facility for setting/getting it -- until then it simply remains initialized to 1.

Don't expect too much from these simple units though, afterall they can only take 1024 samples at up to 1MHz rate on 3 channels.

While stuck working from home during COVID19 quarantine, I needed a logic analyzer to measure timing on a serial port that I suspected might be off.  All I had lying around was a PICkit3 and a few loose PIC chips of various kinds.  So, I endeavored to use the PICkit3, and when I found that sigrok had a PICkit2 driver, I figured the PICkit3, using Microchip's "PICkit3 Programming App and Scripting Tool v3.10", should work with a few minor tweaks.  Microchip's basic logic analyzer GUI app probably would have worked fine for my needs, but I stubbornly wanted to use sigrok -- too much time on my hands, I guess.  By the time I was finished, I had found several areas that needed to be fixed up to get it working properly with the PICkit2 and the PICkit3.

I renamed the driver "microchip-pickit" since it now works with both the 2 and 3 (removed the 2 from the driver name; updated Makefile, etc, and used `git mv` so the diffs would be tracked).  The driver will scan for all Microchip PICkit2 and PICkit3 devices plugged into usb, as well as clones that use the same VID & PID.  The `-d microchip-pickit:conn=` configuration option on the command line will also work, and is checked first, in case a clone exists that does not mimic the OEM VID & PID.  In this case, I added the ability to tell the driver whether the clone mimics a PICkit2 or PICkit3 by adding "#2" or "#3", respectively, for instance, `-d microchip-pickit:conn=VID.PID#3`, which tells the driver that it is working with a PICkit3.  (Note: the driver can not tell if the PICkit3 is flashed with its standard PIC programming firmware, or with the necessary standalone "PICkit3 Programming App and Scripting Tool v3.10" as mentioned above.  So, if it recognizes your PICkit3, but does not seem to work, make sure you have the correct firmware flashed.

I have fully tested the driver with the PICkit3.  I don't have a PICkit2 to test it with, but I am confident that it will work, as there are only a few differences between the two that had to be handled (search for `isPk3` in the code, if interested).

The PICkits are not very relevant, certainly for use as a logic analyzer, but I had a need, and like anything else us geeks do, coding has an aspect of "fun" to it, and satisfaction in "making it work".  I place this here for the community.  I have not checked to see if it meets any coding standards or other rules outlined by this community, but I hope it will stand up.


abraxa:
Hello, thanks for your contribution and sorry for the late feedback!

Unfortunately, the PR is hard to review because there are no individual commits. Did you squash your changes before the push? If so, please push again without the squashing so that we have an easier time reviewing the changes piece by piece.

Also, the driver shouldn't be renamed so we don't break existing user setups/scripts, even if indeed the 2 in the name is less than ideal. New users will likely assume the driver is compatible anyway and just try it, which then shows that it'll "just work" with the PICkit3 as users hoped.

Can you please elaborate why the #2 and #3 are needed? This driver is the only one that would make use of this non-standard convention and if there's a way around it we very much would like to avoid it.

Again, thanks for working with us and hoping to hear from you!

ean365:
Soeren:

No problem, we all have "day jobs" that keep us busy and distracted from this fun stuff.  I've sort of moved on since I worked on this.  If I remember correctly, I did not make intermediate commits.  There was a lot that needed to be fixed or added in many areas of the code to get it working properly with both the pickit2 and pickit3.  I needed to get it working quickly at the time, because I needed the "tool", so I probably did it all at once, tested it all, then checked it in.  So, unfortunately, I understand that it may be difficult to review in small chunks as individual changes, but that was not practical.  I did try to comment it thoroughly.

Although well-intentioned, it was probably presumptuous of me to rename the module from microchip-pickit2 to microchip-pickit.  I can undo that and check it back in.  (Also, some of the difficulty in viewing "changes" may be my renaming of all "pickit2" variables to "pickit"...those can be changed back fairly easily, too...though might be confusing to someone in the future looking at the code -- a bit more confusing distinguishing between pickit2 specific code and pickits3 specific code and common pickit code.)

However, part of my reasoning, and to answer your question regarding the `#2` and `#3` in the command line `-d microchip-pickit:conn=VID.PID#n`...
The pickit2 and pickit3 use very similar protocols, but with some major differences, and there is no easy way to interrogate them to determine whether it is a pickit2 or a pickit3 in order to communicate with it correctly.

If it is a genuine pickit, then you can simply tell by the OEM VID:PID, but if it is a clone or non-standard, and the user tries the command line `-d microchip-pickit:conn=VID.PID` option, I needed a way for the user to also instruct the driver whether to treat it as a pickit2 or pickit3.  Adding the `#2` or `#3` seemed like the most convenient way to do it, and I figured it would not interfere in any way with how any other device driver would treat that option.  If there is a better way to supply this information from the command line, or you think I shouldn't bother handling possible "clones", then I would be happy to implement the better way or remove "clone" handling with the `-d microchip-pickit:conn=` option.

Unfortunately, I don't think that the driver will "just work with the pickit3 as users hoped", since they will have to know to load alternate Microchip supplied firmware (see my OP above) onto their pickit3, and of course if it uses a non-standard VID:PID, then it will only work if they properly instruct the driver whether it is emulating a `#2` or `#3`.

I am happy to work with you as time permits.  Please let me know your preferences.

abraxa:
Hello, after careful consideration it was decided that support for the PICkit family of devices will be removed. While the "make the most of the tools you have when in a bind"-approach is commendable, 1024 samples at 1 MHz is too limiting even in the most desperate situations. Regrettably, this means that this PR won't be included but I hope you'll continue submitting future work - ideally after consulting with us on IRC to align on our efforts.

PR #129, opened 2021-03-25
Faster .sr file save performance.

markrages:
Instead of using the libzip functions (which do all manner of file
copying under the covers), just create a directory while saving.  Then
compress with a call out to /usr/bin/zip.

By default, compress at level "-2".  But the environment variable
"SR_COMPRESSION_FLAG" can be used to select other compression levels, to
trade off speed for size.

If size is a big concern, the best compression is obtained by using
compression flag "-0" (for store-only) and then compressing the
resulting file with something like bzip2.  This exploits redundancy
between the data chunks, compression to less than half the size of
zip -9.

With this patch I can download from my Saleae Logic16 indefinitely at 80
MHz sample rate using sigrok-cli.  This was previously not possible.

### DIFF
--- a/src/output/srzip.c
+++ b/src/output/srzip.c
@@ -65,7 +65,6 @@ static int init(struct sr_output *o, GHashTable *options)
 	outc->filename = g_strdup(o->filename);
 	o->priv = outc;
 
-
 	return SR_OK;
 }
 
@@ -144,7 +143,7 @@ static gboolean write_to_file(const char *dirpath,
 	// in newer glib could just use g_file_set_contents()
 
 	errno = 0;
-	FILE *fd = fopen(path, "w");
+	FILE *fd = fopen(path, "wb");
 
 	if (!fd) {
 		sr_err("Failed to open file %s: %s",
@@ -154,7 +153,7 @@ static gboolean write_to_file(const char *dirpath,
 	}
 
 	if ((1 != fwrite(data, len, 1, fd)) || errno) {
-		sr_err("Failed to open file %s: %s",
+		sr_err("Failed to write to file %s: %s",
 		       path, strerror(errno));
 		g_free(path);
 		return 0;
@@ -711,7 +710,7 @@ static int zip_append_analog_queue(const struct sr_output *o,
 // zip up our directory to a standard "sr" file.
 static gboolean zip_to_srfile(gchar *filename)
 {
-	gboolean ret = 1;
+	gboolean ret = TRUE;
 
 	// "example.sr" --> "example.sr.dir"
 	gchar *dirname = g_strdup_printf("%s.dir", filename);
@@ -724,7 +723,7 @@ static gboolean zip_to_srfile(gchar *filename)
 		sr_err("Failed to rename %s to %s",
 		       filename,
 		       dirname);
-		ret = 0;
+		ret = FALSE;
 		goto cleanup_dir_rename;
 	}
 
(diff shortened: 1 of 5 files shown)


src/output/srzip.c
```diff
@@ -632,6 +706,88 @@ static int zip_append_analog_queue(const struct sr_output *o,
...
+	gchar *zipname = g_build_filename("..",
+	                                  basename,
+	                                  NULL);
+
+	const gchar *compression = g_getenv("SR_COMPRESSION_FLAG");
+	if (!compression) {
+		compression = "-2"; // fastest in my tests.
+	}
+
+	gchar *argv[] = { "/usr/bin/zip",
```
knarfS, line 736: I'm afraid, this is breaking windows support

markrages: Should be fixed now.

jr-oss:
Do you really want to modify srzip output driver?
I would rather have a new output driver with that behavior than risking a regression for other users.
You are using it with sigrok-cli, i.e. it would be no effort for you to use that new output driver.

I also had the idea of storing the data uncompressed to disk to accelerate sigrok-cli.
But my use case is different: I capture data with my scope and save it for later analysis.
PV needs 60s to download the data, while sigrok-cli takes 4 minutes to download and store the data.
Actually PV also needs 3 minutes to store it, so the time sigrok-cli needs is spent mostly in srzip output.
(This increases likelihood of USB buffer overruns with sigrok-cli.)

During my tests, I noticed, that (see data at the end of my comment)
- extracting my captured srzip file to an directory and compressing it with "zip -9" gives roughly the same compressed file size.
  And it takes 3 minutes to compress :-(
- compressing the directory with default compression only needs 15s for the same data

Are you sure, that your gain in speed is only caused by using miniz and avoiding copying files in zlib?
My guess is, that the compression level has a much higher influence that copying files.

If you could change srzip to accept a compression level from command line *and* have zlib obeying it, this might be a good solution.
You could try to patch srzip output driver locally to use store-only method, to see if it is fast enough for your application.
(Just my opinion as a user, not being responsible for any part of sigrok project)


max compression
uncompressed size 430M
compressed size: 24M

$ time zip -9 ../test.sr *
  adding: analog-1-1-1 (deflated 96%)
  ...
  adding: analog-1-20-9 (deflated 94%)
  adding: logic-1-1 (deflated 100%)
  ...
  adding: logic-1-9 (deflated 100%)
  adding: metadata (deflated 31%)
  adding: version (stored 0%)

real    2m57.435s
user    2m57.324s
sys     0m0.080s
  


default compression
uncompressed size 430M
compressed size: 28M

$ time zip ../test.sr *
  adding: analog-1-1-1 (deflated 96%)
  ...
  adding: analog-1-20-9 (deflated 94%)
  adding: logic-1-1 (deflated 100%)
  ...
  adding: logic-1-9 (deflated 100%)
  adding: metadata (deflated 31%)
  adding: version (stored 0%)

real    0m14.150s
user    0m13.986s
sys     0m0.161s


BTW: 80MHz with Saleae logic16, i.e. 160MB/s over USB 2.0?
I would like to know how to do it ;-)

biot:
Sorry it took so long to get reviewed.

I'm not convinced pulling in miniz is the right approach (even less shelling out to `/usr/bin/zip`).

Nevertheless, you and @jr-oss have shown there's much performance to be gained here. I would suggest a call to `zip_set_file_compression()` in `src/output/srzip.c` would be the place to start, driven by a `compression` option, set to a format like `bzip2:9` to pass compression method and level. This breaks nothing, is backward compatible, and may be all that's needed to fix performance.

Krakonos:
Hi! I did some digging here and likely found the reason.

1. By default, compression is on. We can easily set compression level to `store` to get some extra performance at the cost of storage. However, this is NOT the issue here.
2. The core issue is in `zip_append` and `zip_append_analog`. As each chunk is added to the archive, they call `zip_open`/`zip_close`. This leads to the archive to be kept up to date. Unfortunately, the library does not modify the archive in place, but rather creates a copy of the archive `filename.sr.randomstring.part` compress all the files into there (potentially also recompressing files, I'm not sure about this, but I've seen a bit too much compression action for my liking).

I think the best solution here is to keep the archive open until `SR_DF_END`. This shouldn't be hard, though some extra memory will be required to store the intermediate data. Streaming this directly to a filesystem would be better, but I don't see a decent way to do it (and I dislike the miniz solution TBH).

PR #132, opened 2021-04-08
build: Correct linking of libsigrok.la.

orbea:
When building libsigrok with slibtool (https://dev.midipix.org/cross/slibtool) it fails.
```
rdlibtool --tag=CC --silent --mode=link gcc -std=c99 -fvisibility=hidden -Wall -Wextra -Wmissing-prototypes -I/usr/include/glib-2.0 -I/usr/lib64/glib-2.0/include -I/usr/include/libusb-1.0 -I/usr/lib64/libffi/include -pthread -I/usr/include/libmount -I/usr/include/blkid -g -O2 -o src/libdrivers.o src/libdrivers_head.la src/libdrivers.la src/libdrivers_tail.la

gcc: error: src/.libs/libdrivers.o: No such file or directory
gcc: error: missing filename after ‘-o’
rdlibtool: exec error upon slbt_exec_link_create_executable(), line 1731: (see child process error messages).
rdlibtool: < returned to > slbt_exec_link(), line 2051.
make: *** [Makefile:5421: src/libdrivers.o] Error 2
```
This shows two problems with `Makefile.am`.

1. It should not manually overwrite the make targets for `src/libdrivers.o` and `src/libdrivers.lo`, just adding `src/libdrivers.la` to `noinst_LTLIBRARIES` is enough.
2. The build should ideally not link with `.lo` files, but with `.la` files. It is also enough to add the required `.la` files to `LIBADD` for libraries or `LDADD` for programs.

GNU libtool does not fail because it is far more permissive even to the point of hiding real issues...

Also see this downstream issue: https://bugs.gentoo.org/780840


jr-oss:
There have been lengthy discussions and tests about linking libsigrok.la:
1. https://sigrok.org/bugzilla/show_bug.cgi?id=1433
2. https://sourceforge.net/p/sigrok/mailman/message/37233640/

orbea:
I don't know about the LTO issue, but the current Makefile.am is just wrong, unfortunately GNU libtool is an extremely buggy unmaintained shell script which accepts all sorts of incorrect input without complaint...

gsigh:
Your response strongly suggests that you are not interested in understanding the context of what you change, nor in providing something that considers the project's needs when it's supposed to work in rather different environments across several platforms. You seem to scratch an itch of yours while you completely disregard the communication partner's position. Which dramatically reduces the probability of your change getting accepted.

PR #133, opened 2021-04-12
Update api.c

danselmi:
Fix frequency range fro SME06 according to:
https://scdn.rohde-schwarz.com/ur/pws/dl_downloads/dl_common_library/dl_manuals/gb_1/s/sme_1/sme_14e.pdf


danselmi:
should be replaced py PR140

PR #136, opened 2021-05-04
Build: Remove libusb_os_handle requirement for Windows

KyleJ61782:
This PR removes the build requirement as it's not needed to successfully build in MinGW.


gsigh:
This approach is overly eager. It _may_ unbreak the Windows build with newer libraries, but _does_ break them when the older library is provided (as is the case in all existing setups which don't upgrade the libusb dependency).

See commit db39d9a271e3 which makes the data type's absence non-fatal on Windows, but should keep all other configurations unaffected. Feel free to follow up or reopen should current mainline not work for you as this submission of yours did.

PR #140, opened 2021-05-08
Add drivers for R&S SMC100A, SMB100A and SMBV100A

danselmi:
Can someone please review this driver?

It adds support for some signal generators form R&S.

Is there anything else I can do to get this into head?


knarfS (commented):
Some additional notes:

- Could this driver fit into the R&S SME driver?
- You should create 2 commits: one for the skeleton and one for your driver implementation.
See https://sigrok.org/gitweb/?p=libsigrok.git;a=blob;f=HACKING;hb=HEAD:
  > That means your patch submission later will consist of at least two patches: the initial one adding the skeleton driver, and one or more additional patches that actually implement the respective driver code.
- I'm curious, if this driver will work with SmuView. As there is no acquisition, a mutex is not needed.
I'll add the new device type to SmuView later that day and let you know :)

danselmi:
Thanks for your feedback.

I tried to address you comments in 4703bfd and 84e3ca3.
Except the step-size thing.

In 13e05ce I added the support in the R&S SME driver. This would make the other commits obsolete.
In this solution the step-size is also correct.
I haven't tested these changes yet. Will do this on Thursday.
I think it is better to have them all in the same driver. But the driver name seems wrong now.

danselmi:
I was able to test the sme-0x driver with the following devices:
SME-03, SMB100A, SMBV100A and SMC100A.
All worked as expected.

All except the SME-03 also worked with SmuView-git-d926030. This only worked with sigrok-cli. Probably because the baudrate can not be configured.

I would prefer to not use the smx100 driver but change sme-0x according to this branch. So I have not tested the smx100 driver anymore.

biot (changes requested):
Code looks great, very cool how you drop the hardcoded min/max power and frequency and fetch it dynamically from the hardware instead.

Can you just squash the second commit?

danselmi:
The PRESET and RESET_STATUS commands are not called for sme0x devices. This is due to backward compatibility to the old implementation. I think this is wrong.

On the other hand I can imagine use cases where you configure the the signal "by hand" and use sigrok only for sweeps.

What do you think?

src/hardware/rohde-schwarz-sme-0x/protocol.c
```diff
@@ -24,66 +25,368 @@
...
+
+	devc->clk_source_idx = idx;
+
+	return sr_scpi_send(sdi->conn, (idx == 0) ?
+					devc->model_config->commands[RS_CMD_SET_CLK_SRC_INT] :
+					devc->model_config->commands[RS_CMD_SET_CLK_SRC_EXT]);
+}
+
+SR_PRIV int rs_sme0x_get_minmax_freq(const struct sr_dev_inst *sdi,
+									double *min_freq, double *max_freq)
```
biot, line 367: That's hard to read, just indent it with one tab (same for `rs_sme0x_get_minmax_power`).

biot (changes requested):
> I think this is wrong.

Then you should fix it. Correct handling of the hardware should take precedence over (purely theoretical) reliance on broken code.

Also just the one coding style fix left.

biot (approved):

PR #143, opened 2021-05-25
add udev rule for Agilent 34460A USBTMC

sylvainpelissier:


gsigh:
Merged in 1b7d49db6444. Thank you.

PR #144, opened 2021-06-09
link-mso19: fixup old code to actually work and enable hardware

pkasemir:


gsigh:
Would suggest to unbreak compilation first and hook up to the build afterwards. Increases maintainability (think bisect). Adding features or fixing issues beyond mere compilation issues can be done after the compilation starts working again, optionally before or after hooking up to the build (ideally when covered by the build).

Please check compilation with different library configurations (satisfied or unmet dependencies, especially serial communication). It appears that the previous mainline implementation did redundant and highly platform specific things with USB enumeration, the updated version would strictly depend on libserialport? At least declare that dependency to avoid build errors in environments where the dependency is not available. Ideally avoid the libsp dependency because serial communication can optionally run on other physical transports, too.

pkasemir:
@gsigh I've addressed your comments.

I've force pushed to rebase onto latest master. (saved old commits in https://github.com/pkasemir/libsigrok/tree/mso19_save) 
There were no conflicts, but I did change the order of the 1st 3 commits to only enable the hardware after fixing compiler errors. Now _all_ commits will build. (this was a good idea, just forgot since I needed to enable build to fix it first) 

I also added the dependency on libserialport.

I've tested and found that the HW still works just fine after the rebase.

I understand your desire to not use libsp as  dependency since there are possible other transports, but in this case I think it really does make sense. 
* The libserialport seems to be a core library for libsigrok (the scripts from git://sigrok.org/sigrok-util use libserialport for **all** OSes, so I find it an adequate dependency) 
* Also I'm pretty sure bluetooth libs would not function as `serial_comm` for this device. 
* The link-mso19 device strangely puts HW config values in the USB iSerial, so I need to use that otherwise the calibration could be wrong giving slightly incorrect voltages. I suppose we could use normal serial functions to find it and then use libusb to try and gather iSerial, but that seems pointless as libsp already has that function **and** then we would need serial_comm and libusb dependencies.  libsp seems to be the best catch all to allow the device to function on all OSes.

mkj:
Thumbs up from me for merging this. I've tested building on Ubuntu 22.04 and it works well, at least testing with LA triggers.
Thanks @pkasemir

pkasemir:
Force push to rebase onto latest `master` branch (no conflicts and no changes)

I've rebuilt and tested and still works as expected.

pkasemir:
@gsigh can you please look again? I'm having zero feedback from mailing list nor IRC even with numerous attempts asking for review.

PeterMortensen:
Can you **_schedule_** a time on IRC? One hour is not enough response time. For instance, within 1400 hours - 2300 hours UTC. Likely the best time is 2000 hours UTC.

src/hardware/link-mso19/protocol.c
```diff
@@ -66,12 +66,20 @@ SR_PRIV int mso_send_control_message(struct sr_serial_dev_inst *serial,
...
+	return mso_trans(REG_CTL2, devc->ctlbase2 | BITS_CTL2_BANK(bank));
+}
+
 SR_PRIV int mso_configure_trigger(const struct sr_dev_inst *sdi)
 {
 	struct dev_context *devc = sdi->priv;
 	uint16_t threshold_value = mso_calc_raw_from_mv(devc);
 
-	threshold_value = 0x153C;
+	threshold_value = 0x200;
```
fenugrec: magic constant

src/hardware/link-mso19/api.c
```diff
@@ -267,7 +276,10 @@ static int config_set(uint32_t key, GVariant *data,
 		trigger_pos = (int)pos;
 		devc->trigger_holdoff[0] = trigger_pos & 0xff;
 		break;
-	case SR_CONF_RLE:
+	case SR_CONF_COUPLING:
+		if (!cg_is_analog(cg) || (idx = std_str_idx(data, ARRAY_AND_SIZE(coupling))) < 0)
```
fenugrec, line 497: a lot of stuff going on in this conditional, suggest splitting assignment and conditional on 2 lines

src/hardware/link-mso19/api.c
```diff
@@ -262,10 +333,33 @@ static int config_set(uint32_t key, GVariant *data,
 		break;
 	case SR_CONF_CAPTURE_RATIO:
 		break;
-	case SR_CONF_TRIGGER_SLOPE:
-		if ((idx = std_str_idx(data, ARRAY_AND_SIZE(trigger_slopes))) < 0)
+	case SR_CONF_TRIGGER_SOURCE:
+		if ((idx = std_str_idx(data, ARRAY_AND_SIZE(trigger_sources))) < 0)
```
fenugrec, line 447: another asignment-inside-conditional

src/hardware/link-mso19/api.c
```diff
@@ -262,10 +333,33 @@ static int config_set(uint32_t key, GVariant *data,
...
+			break;
+		case TRIGGER_SOURCE_LA:
+			TRIG_UPDATE_SRC(devc->ctltrig, TRIG_SRC_LA);
+			break;
+		}
+		mso_update_trigger_slope(devc);
+		break;
+	case SR_CONF_TRIGGER_SLOPE:
+		if (cg_is_analog(cg)) {
+			if ((idx = std_str_idx(data, ARRAY_AND_SIZE(dso_trigger_slopes))) < 0)
```
fenugrec, line 462: asignment-inside-conditional

src/hardware/link-mso19/api.c
```diff
@@ -262,10 +333,33 @@ static int config_set(uint32_t key, GVariant *data,
...
+				return SR_ERR_ARG;
+			devc->dso_trigger_slope = idx;
+		} else if (cg_is_digital(cg)) {
+			if ((idx = std_str_idx(data, ARRAY_AND_SIZE(la_trigger_slopes))) < 0)
```
fenugrec, line 466: asignment-inside-conditional

src/hardware/link-mso19/protocol.c
```diff
@@ -449,40 +414,33 @@ SR_PRIV int mso_receive_data(int fd, int revents, void *cb_data)
...
-			continue;
-
-		int channel_bit = 1 << (ch->index);
-		if (!(ch->trigger))
-			continue;
-
-		devc->use_trigger = TRUE;
-		//Configure trigger mask and value.
-		for (tc = ch->trigger; *tc; tc++) {
+	if (!(trigger = sr_session_trigger_get(sdi->session)))
```
fenugrec, line 460: asignment-inside-conditional

src/hardware/link-mso19/api.c
```diff
@@ -134,6 +133,30 @@ static void mso_limit_trigger_level(struct dev_context *devc)
 	sr_info("Adjusted dso trigger level to %f", devc->dso_trigger_adjusted);
 }
 
+static void mso_update_trigger_pos(struct dev_context *devc)
+{
+	int pos = (devc->horiz_triggerpos * MSO_NUM_SAMPLES) + 0.5;
+	uint16_t sign_bit = TRIG_POS_IS_POSATIVE;
```
fenugrec, line 163: typo in _POSATIVE

src/hardware/link-mso19/api.c
```diff
@@ -92,6 +93,24 @@ static const int32_t trigger_matches[] = {
...
+	"1.2V Logic",
+	"1.5V Logic",
+	"1.8V Logic",
+	"2.5V Logic",
+	"3.0V Logic",
+	"3.3V/5.0V Logic",
+};
+
+static const uint16_t logic_threshold_values[] = {
+	0x600,
```
fenugrec, line 107: is there a way to write this that exposes whatever unit this is in ? not sure if  `600*256/100`  would integer promotion /cast issues, but maybe at least a comment that mentions the scaling

src/hardware/link-mso19/api.c
```diff
@@ -140,8 +141,12 @@ static void mso_update_trigger_slope(struct dev_context *devc)
 static void mso_limit_trigger_level(struct dev_context *devc)
 {
 	double max_level = 2.0 * devc->dso_probe_factor;
-	if (devc->dso_trigger_level < -max_level) {
-		devc->dso_trigger_adjusted = -max_level;
+	double min_level = -max_level;
+	max_level -= devc->dso_offset_adjusted;
+	min_level -= devc->dso_offset_adjusted;
```
fenugrec, line 146: isn't this going to set min_level to something outside of the interval [-max, +max] ?

src/hardware/link-mso19/protocol.h
```diff
@@ -206,6 +209,7 @@ enum {
 
 /* bits - REG_DAC */
 enum {
+	DAC_DSO_VALUE_MASK =	0xfff,
 	DAC_SELECT_DSO =	0 << 15,
 	DAC_SELECT_LA =		1 << 15,
```
fenugrec, line 214: is it OK that the DAC_SELECT_* enum values are not covered by DAC_DSO_VALUE_MASK ? i.e. 0xfff is 12 bits, not 15

fenugrec (commented):
I do not have this hardware; I am neither a maintainer nor even a regular sigrok dev, so take my comments with a grain of salt. And please be very patient as this project is severely understaffed these days.

src/hardware/link-mso19/protocol.c
```diff
@@ -24,38 +25,34 @@
 
 /* serial protocol */
 #define mso_trans(a, v) \
-	(((v) & 0x3f) | (((v) & 0xc0) << 6) | (((a) & 0xf) << 8) | \
+	g_htons(((v) & 0x3f) | (((v) & 0xc0) << 6) | (((a) & 0xf) << 8) | \
```
fenugrec, line 28: I believe there is a clearer alternative than g_htons for managing endianness. Ran out of time to look it up, sorry

src/hardware/link-mso19/protocol.c
```diff
@@ -349,15 +350,23 @@ SR_PRIV int mso_receive_data(int fd, int revents, void *cb_data)
...
 
+	uint8_t in[1024];
+	size_t s;
+
+	(void)fd;
 	(void)revents;
 
-	uint8_t in[1024];
-	size_t s = serial_read(devc->serial, in, sizeof(in));
+	if (!(sdi = cb_data))
```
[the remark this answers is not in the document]
pkasemir: This is consistent with all the other hardware. Half of them do this exact line of code.

fenugrec:
Also, something else that popped up on IRC after you left yesterday : there should be 

- clear mentions that this driver has a hard dependency on libserialport
- corresponding adjustments to the build system

src/hardware/link-mso19/protocol.c
fenugrec: that would be `read_be16` and similar funcs

src/hardware/link-mso19/protocol.c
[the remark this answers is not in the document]
fenugrec: > This is consistent with all the other hardware. Half of them do this exact line of code.

You are correct, but but it makes reviewing & future maintenance more difficult - I need to pause and double-check to see if it was supposed to be "==" , and if not, figure out the logic.

Discussions on IRC that you missed yesterday were in that direction too ; it's an easy fix that only makes the code more readable.

knarfS:
I'd like to point to the [Linux kernel coding style guid](https://www.kernel.org/doc/html/latest/process/coding-style.html), which is also the style guide for the sigrok project (see 1st paragraph in [HACKING](https://github.com/sigrokproject/libsigrok/blob/master/HACKING)).
In particular, the 80 character limit per line.

pkasemir:
Sorry for force push, just removed the merge commit with `master` that github automatically added.

src/hardware/link-mso19/protocol.c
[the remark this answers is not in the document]
pkasemir: However, this is code directly from the HACKING file 

> Adding a new hardware driver
> ----------------------------
> 
> The simple, scripted way (recommended):
> ---------------------------------------
> 
> Use the 'new-driver' script from the sigrok-util repo:
> 
>  $ git clone git://sigrok.org/sigrok-util
>  $ cd sigrok-util/source
>  $ ./new-driver "Tondaj SL-814"

See https://github.com/sigrokproject/sigrok-util/blob/master/source/drv-protocol.c#L30

If this is desired change, then the new code should also reflect this.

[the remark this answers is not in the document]
fenugrec: Heh, I wasn't aware of that script !  Here... https://github.com/sigrokproject/sigrok-util/pull/14

src/hardware/link-mso19/protocol.h
pkasemir: Yes it is correct. The value part is 12 bits and the select bit is bit15.

src/hardware/link-mso19/api.c
pkasemir: This code is correct. Think of it like a sliding window. the MSO can slide the window up or down based on DAC value.

So normally, you might do a [min, max] of [-2.0, 2.0], but you can also do [-1.0, 3.0] or [0.0, 4.0]

pkasemir:
Force push to resolve many of the pointed out issues. Please check again @fenugrec

src/hardware/link-mso19/api.c
fenugrec: Ok

src/hardware/link-mso19/protocol.h
fenugrec: ok

src/hardware/link-mso19/protocol.c
```diff
@@ -56,7 +53,7 @@ SR_PRIV int mso_send_control_message(struct sr_serial_dev_inst *serial,
 
 	w = 0;
 	while (w < s) {
-		ret = serial_write(serial, buf + w, s - w);
+		ret = serial_write_blocking(serial, buf + w, s - w, 10);
```
fenugrec, line 55: I think this is the type of commit that gsi objects to - the commit line is "remove stale code to get driver to compile ", but here is a change that doesn't really fit with the message.

src/hardware/link-mso19/protocol.c
```diff
@@ -199,7 +199,7 @@ SR_PRIV int mso_parse_serial(const char *iSerial, const char *iProduct,
 		u3 = 500;
 	devc->vbit = ((double)u1) / 10000.0 / 1000.0;
 	devc->dac_offset = u2;
-	devc->offset_range = u3;
+	devc->offset_vbit = 3.0 / u3;
```
fenugrec, line 202: Initially I had missed this "3.0" number - where is that from ?

fenugrec (commented):
Much better IMO; I haven't looked at the 80-char width limit though.
I also agree that at least one of the file's headers should probably explain why it depends on libserialport for parsing features ?

src/hardware/link-mso19/protocol.c
pkasemir: I trying to agree with you, but simply can't. `serial_write` doesn't exist anymore (and thus is stale). So I use the appropriate replacement for the stale API.

src/hardware/link-mso19/protocol.c
pkasemir: I'm not exactly sure exactly. It is basically taken from this other project https://github.com/tkrmnz/mso19fcgi/blob/master/main.c#L332

It's been a while since I last looked at that code, so it is a bit fuzzy, but I think the off by factor of 1000 is probably due to mV to V conversion somewhere differing in their code vs ours.

I'm pretty sure I tested it and it worked a long time ago.

pkasemir:
force push to simply change bf77677b39628ddedc7a757f2783f88066b4823e to bf77677b39628ddedc7a757f2783f88066b4823e in both comment and commit message

Only change `libserial` -> `libserialport`

pkasemir:
> I also agree that at least one of the file's headers should probably explain why it depends on libserialport for parsing features ?

@fenugrec how many placed do you need this, IMO, rather obvious dependency? It is already in:
1. configure.ac
2. the commit message for c7f8020c2ca37640e7efc010228d546728e0d654
3. a comment in `api.c` right before the call to `sp_get_port_usb_serial()`

> - corresponding adjustments to the build system

What do you mean by this? I've change the build system to build the code, and all the commits compile. What more do you need than that?

pkasemir:
@fenugrec I typically let the commenter resolve their posted issues. If you wouldn't mind resolving things, that will clean up the PR.  (or I can close them too)

pkasemir:
As for the 80 line limit? how hard of a requirement is that? Most of the lines that are over the limit are just a few chars too long. 
 Some of the lines would look worse if split to fit 80 chars (like log strings)

knarfS:
> As for the 80 line limit? how hard of a requirement is that? Most of the lines that are over the limit are just a few chars too long. Some of the lines would look worse if split to fit 80 chars (like log strings)

Have a look at `mso_calc_trigger_threshold()`. Or the two comments in `mso_configure_channels()`.
One character should be ok. Also longer definitions for `devopts[]`, `devopts_cg_analog[]`, etc in `api.c` should be ok.

But I'd see that as an requirement. From the linux kernel coding guidlines:
```
2) Breaking long lines and strings

Coding style is all about readability and maintainability using commonly available tools.

The limit on the length of lines is 80 columns and this is a strongly preferred limit.

Statements longer than 80 columns will be broken into sensible chunks, unless exceeding 80 columns significantly increases readability and does not hide information. Descendants are always substantially shorter than the parent and are placed substantially to the right. The same applies to function headers with a long argument list. However, never break user-visible strings such as printk messages, because that breaks the ability to grep for them.
```

> There are also a ton of lines longer than 80 chars already in libsigrok.

Yeah, but that doesn't make them right ;)

fenugrec:
> how many placed do you need this, IMO, rather obvious dependency

I personally think it's fine, but was echoing some comments from IRC that you may have missed.

> I typically let the commenter resolve their posted issues. If you wouldn't mind resolving things,

I can't ! Usually there's a button "Resolve conversation" but seems like I don't have sufficient privs.


To restate - I'm just trying to help get this patchset to as high a quality as possible, in the hope that it becomes "good enough" to be picked up by the only currently-active dev with commit access, who also has extremely limited time to engage in discussions like we have. I.e. trying to make a sale, in a sense.

src/hardware/link-mso19/protocol.c
fenugrec: I think that's probably the last "magic number" that would be nice to get rid of. TBH I haven't gone back to look at what `offset_vbit` actually does, so maybe this is a non-issue.

src/hardware/link-mso19/api.c
```diff
@@ -53,124 +63,265 @@ static const uint64_t samplerates[] = {
...
+			TRIG_UPDATE_EDGE(devc->ctltrig, TRIG_EDGE_T_F);
+			break;
+		}
+		break;
+	}
+}
+
+static void mso_limit_trigger_level(struct dev_context *devc)
+{
+	double max_level = 2.0 * devc->dso_probe_factor;
```
knarfS, line 144: As a convention in this project, all variables used in a function are declared at the start of that function first and used later.

src/hardware/link-mso19/api.c
```diff
@@ -53,124 +63,265 @@ static const uint64_t samplerates[] = {
...
+	serial_num = sp_get_port_usb_serial(port);
+	if (mso_parse_serial(serial_num, product, devc) != SR_OK) {
+		sr_err("Invalid serial: %s.", serial_num);
+		g_free(devc);
+		return devices;
+	}
+	sprintf(hwrev, "r%d", devc->hwrev);
+	devc->ctlbase1 = BIT_CTL1_ADC_ENABLE;
+	TRIG_UPDATE_OUT(devc->ctltrig, TRIG_OUT_TRIGGER);
+	TRIG_UPDATE_SRC(devc->ctltrig, TRIG_SRC_DSO);
```
knarfS, line 245: Macros: see  above

src/hardware/link-mso19/api.c
```diff
@@ -53,124 +63,265 @@ static const uint64_t samplerates[] = {
...
+		devc->dso_trigger_adjusted = max_level;
+	} else if (devc->dso_trigger_level != devc->dso_trigger_adjusted){
+		devc->dso_trigger_adjusted = devc->dso_trigger_level;
+	} else {
+		return;
+	}
+	sr_info("Adjusted dso trigger level to %f", devc->dso_trigger_adjusted);
+}
+
+static void mso_update_trigger_pos(struct dev_context *devc)
```
knarfS, line 161: Smth like `mso_get_trigger_pos(...)` that returns the calculated position (and not setting it) would be better.

Same for the two functions above.

src/hardware/link-mso19/protocol.c
```diff
@@ -24,38 +25,34 @@
...
 static const char mso_head[] = { 0x40, 0x4c, 0x44, 0x53, 0x7e };
 static const char mso_foot[] = { 0x7e };
 
-SR_PRIV int mso_send_control_message(struct sr_serial_dev_inst *serial,
-				     uint16_t payload[], int n)
+static int mso_send_control_message(struct sr_serial_dev_inst *serial,
+				    uint16_t payload[], int n)
 {
-	int i, w, ret, s = n * 2 + sizeof(mso_head) + sizeof(mso_foot);
+	int w, ret;
```
knarfS, line 37: I'd prefer more  speaking variable names

src/hardware/link-mso19/protocol.h
```diff
@@ -25,34 +26,20 @@
...
-
-enum trigger_slopes {
-	SLOPE_POSITIVE = 0,
-	SLOPE_NEGATIVE,
-};
+#define MSO_NUM_SAMPLES		1024
+#define MSO_NUM_LOGIC_CHANNELS	8
+
+#define cg_is_digital(cg) (cg && cg->name[0] == 'L')
+#define cg_is_analog(cg) (cg && cg->name[0] == 'D')
```
knarfS, line 42: Macro names should be upper case

src/hardware/link-mso19/protocol.c
pkasemir: I added a comment for it, which is removed in a later commit anyway so doesn't matter anyway.

src/hardware/link-mso19/api.c
pkasemir: I added a comment for it. These values come from USB captures and could be considered to have some constant times mV value, but it seems that would complicate things even more to try and explain it. So I think a comment will work fine.

src/hardware/link-mso19/protocol.c
```diff
@@ -222,17 +191,17 @@ SR_PRIV int mso_parse_serial(const char *iSerial, const char *iProduct,
 	if (iSerial[0] != '4' || sscanf(iSerial, "%5u%3u%3u%1u%1u%6u",
```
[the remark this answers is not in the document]
pkasemir: Sure, but I didn't write this line of code... I didn't sign up to rewrite the entire driver to match coding style. I'd prefer to limit my changes to make the driver functional and leave anything that already works unchanged.

src/hardware/link-mso19/protocol.c
pkasemir: I didn't write this code, so I left the variable names as they were originally written.

src/hardware/link-mso19/protocol.c
pkasemir: @knarfS The existing code for mso-19 is riddled with this type of calculation. I tried to backward engineer and make clear everything I could... If having 100% knowledge of the protocol is a requirement for you to merge this code, then I am going to give up and abandon all hope.

If you are willing to budge a little on your perfect ideals to let this one slide, then I can fix up all your other comments and hopefully get this hardware working for other folks who would like to use it. I can at least make it an aptly named macro if that helps.

It would be sad after putting all the effort I have into this code to just throw it all away.

pkasemir: > > It is basically taken from this other project https://github.com/tkrmnz/mso19fcgi/blob/master/main.c#L332
>
> That makes it even worse: I don't know how much code is copied from there

I don't know about any of the original code, but **none of the code I wrote is copied**. Yes, some of the calculations are based on the other project because I have no idea where else to get those types of calculations... Besides, that project is already linked to in the mso-19 wiki (which is how I found it in the first place).

knarfS: I'm totally aware that you didn't copied that code, but when there is code in here, that was taken from that other project, then there should be a note with a link at the top of this file and probably also in the commit message. A link in the wiki is obviously not enough.

knarfS:
> If you are willing to budge a little on your perfect ideals to let this one slide, then I can fix up all your other comments and hopefully get this hardware working for other folks who would like to use it. I can at least make it an aptly named macro if that helps.

@pkasemir I'm not here to trade, nor are this perfect ideals. I'm not here to discuss whether or not the project guidelines and requirements must be met.
I'm here to help get this PR merged and to catch some of the obvious and ugly things that will probably be pointed out anyway. I'm here to help improve the overall code quality and point out the coding guidelines of this project.

I'm only a user and contributor to libsigrok and am not in a position to decide if this PR can be merged or not. I can only make suggestions.

If there is code full with magic numbers that nobody seems to understand, at least put them at the beginning of that function and add a comment to make it clear. Add comments about what units parts of the function works with (e.g. `devc->vbit = ((double)u1) / 10000.0 / 1000.0;`). This way the developer who comes after you doesn't have to figure it out again.

When you are refactoring and fixing a driver and rewrite it almost completely, why wouldn't you fix parts of the code that is only 2 or 3 lines above? Regardless of whether you wrote it or not.

src/hardware/link-mso19/protocol.h
```diff
@@ -158,12 +159,70 @@ SR_PRIV void stop_acquisition(const struct sr_dev_inst *sdi);
...
+	TRIG_SRC_SPI =		4 << 5,
+	TRIG_SRC_I2C =		5 << 5,
+	TRIG_SRC_LA =		7 << 5,
+	TRIG_SRC_MASK =		7 << 5,
+};
+#define TRIG_UPDATE_MASK(reg, val, mask) reg = (((reg) & ~(mask)) | ((val) & (mask)))
+#define TRIG_UPDATE_THRESH_MSB(reg, val) TRIG_UPDATE_MASK((reg), (val) >> 8, TRIG_THRESH_MSB_MASK)
+#define TRIG_UPDATE_EDGE(reg, val)	TRIG_UPDATE_MASK((reg), (val), TRIG_EDGE_MASK)
+#define TRIG_UPDATE_OUT(reg, val)	TRIG_UPDATE_MASK((reg), (val), TRIG_OUT_MASK)
+#define TRIG_UPDATE_SRC(reg, val)	TRIG_UPDATE_MASK((reg), (val), TRIG_SRC_MASK)
```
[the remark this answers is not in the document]
knarfS: I'd like to suggest to remove most of the macros from `protocol.h`: They are not type save, hide functionality from the developer and are hard to maintain.
The `TRIG_UPDATE_MASK` macro is in my opinion a no go, because it additionally contains an assignment to one of its parameters.

pkasemir:
Hi @knarfS I've pushed up some more commits addressing all your comments. If you have the resolve button on them, that would be great to show which ones are done.

I've altered the _update_ style functions, most of them have now become _setter_ style functions. The reason I used _update_ style in the first place was because one setting cascades through and affects many others (as in the register cache value gets updated, or the min/max limits on the voltage range change and could thus move the trigger level). So hopefully this will satisfy your coding style.

pkasemir:
@knarfS and @fenugrec thanks very much for your comments. I think I've dealt with most or all of them, so please take a look at the latest commits.

What is the norm for code review commits? do they get squashed to a single commit? The entire PR squash to a single commit? Obviously this is now a lot of commits (including code review rework) which I understand some folks may not want, but it would be a real pain for me to absorb all the code review commits into the originals.

fenugrec:
> What is the norm for code review commits? 

No idea, but I would recommend against squashing the whole PR as one commit - that would be completely unreviewable, I certainly wouldn't even look at it.

>  it would be a real pain for me to absorb all the code review commits into the originals.

Agreed, but it may be necessary (I am not placed to judge), depending on what they are - I haven't looked at this new set. Sometimes you can get lucky with an interactive rebase ?  just reordering the "fixup" commits  in the correct places and marking them as F ...

pkasemir:
FYI: the summary of changes for all the code review commits
4baf10fc33d1ca6e6b5e26518eccfd6ae9faaf65..a6b89df6ab0df875c8c8368f8374501f254f4609
```
 src/hardware/link-mso19/api.c      | 174 ++++++++++++++++++++++++++++++++++----------------------------
 src/hardware/link-mso19/protocol.c | 132 +++++++++++++++++++++++++++++++----------------
 src/hardware/link-mso19/protocol.h |  21 +++-----
 3 files changed, 190 insertions(+), 137 deletions(-)
```

fenugrec (approved):
LGTM
only saw 1 typo at quick glance

raphCode:
I am not really a C developer or a sigrok user (at the moment), I just want to chime in because I read the discussion on the mailing list.

I want to point out that in order to review something, I should be easy to understand the changes and the reasoning for them. In the case of a git-managed project, information why the change is necessary and how it was performed is usually embedded in the commit messages.
Think of it as a form of communication: The commit message should establish a mental context around problem the commit tries to solve and explain what and why something was done the way it was done.
This does not only help reviewers, but also maintainers and anyone who touches the code again, probably years in the future. It might even help your future self: you mentioned your memory about some code snippet was "fuzzy". This is where commit messages can provide valuable context. Familiarize yourself with the `git blame` command to see why this is important.
Ask yourself: Can you show this commit to a random person / software developer and have them understand the change and why it is important?

There are a few resources around good commit messages, but the one article everyone comes back to is:
https://cbea.ms/git-commit/
I even found this recent academic paper (but haven't read it thoroughly): https://arxiv.org/abs/2202.02974

---

Applying the above to this PR: I checked some of the commits and honestly, I couldn't understand a thing. This may be in part because of I am not very deep in this project and its C code, but also because of the relatively low quality of the commits itself.
I could pick some examples, but it doesn't matter: Either the commit message body is missing completely, or it contains a few lines which don't explain anything in diffs touch dozen or even hundreds of lines of code.

I can't really comment on the actual code itself, but adhering to good commit messages in general also improves code quality:
One needs to think more about the reasoning behind changes: what did the old code do and how? This also forces an understanding of the existing code. Try the mentioned `git blame` on parts of the code you don't understand. What was the problem with the old code? How can it be solved? What are the tradeoffs of your solution?

pkasemir:
@raphCode the article you linked is quite valuable. After reading it I would actually say most of my commits follow the paradigm set there. Yes there may be a few places where I left out some detail, but if you are not a C programmer, then you shouldn't criticize the commit messages for a C project. A lot of the explanation in the changes are self-explanatory by the code diff, and the commit message is just a summary of what changed.

As for commenting on why it changed and why I chose to do it a certain way. I found in unnecessary because the existing code is not even compiled in `master`. I more or less considered my efforts for these code changes as the initial hardware implementation. Explaining why I do everything I do is a waste of time, especially when in most cases the code does that for me.

raphCode:
I know, nobody wants to hear criticism, but still hoped for a different answer.

Anyways, I wish you best of luck getting this merged as well as progress on your ability for self-reflection and to take criticism :)

pkasemir:
@knarfS Hello again. Any chance you could review my updates?

pkasemir:
@knarfS Are you there? I would appreciate if you could look at my fixup commits.

knarfS:
@pkasemir Sorry for the late reply. I'll try to find some time later this week.

mkj:
I'd suggest that this is excessive review. It's not going to break anything outside of the specific hardware, and the hardware is unusable without the patch.

pkasemir:
I would love for this to get merged. But I've completely given up. I tried
tremendously hard to please the other contributors and get this approved,
but I don't have the time anymore.

abraxa:
I have merged your changes, @pkasemir: https://sigrok.org/gitaction/libsigrok.git/e938e2d..1b12204

Thanks for patiently working with us and making sure we understand what you're doing and why. We don't aim to be pedantic just to discourage people but because we've been bit by not being pedantic before, so I hope you understand. The delay in merge is definitely unfortunate and it'll be faster in the future. Thanks for contributing and helping out other users of this driver!

PR #145, opened 2021-06-13
Overhaul FTDI-LA driver: no more dropped samples

tchebb:
The FTDI-LA driver is currently very flaky and, as a result, tends to be treated as a toy rather than a serious driver. However, after lots of investigation and experimentation, I discovered that the perceived flakiness is almost certainly coming almost entirely from two sources:
1. Genuine FT232R chips have an [erratum](https://ftdichip.com/wp-content/uploads/2020/08/TN_120_FT232R-Errata-Technical-Note.pdf) in all silicon revisions which make bitbang timing extremely unstable for both reads and writes. (Ironically, some counterfeit FT232R clones seem to work properly.) As such, genuine FT232R parts are totally unsuitable for use as a logic analyzer, and yet FTDI-LA still claims to support them without so much as a disclaimer.
2. FTDI-LA, via libftdi, currently uses blocking libusb APIs to poll the chip for samples, meaning that there are long periods when the sigrok thread is busy with something else during which the USB bus is completely idle. This leads to periodic device-side buffer overruns at even moderate sample rates.
This pull request in its current form completely fixes the second of these problems by calling directly into libusb's asynchronous API, just like fx2lafw does. I can now reliably capture from an FT2232H at 15MHz (the maximum bitbang frequency) with no dropped samples. Additionally, this PR adds support for capturing from channels other than A on multi-channel devices like the FT2232H and FT4232H.

I have not yet addressed the first problem in this PR, nor have I completed as rigorous testing as I'd like, hence why it's a draft. However, all the hard parts are done and I'd appreciate comments. The only significant additions I'll likely make are better sample rate constraints, since you can currently ask for a rate that the chip does not honor, and a big warning in the log when an FT232R is detected.


src/hardware/ftdi-la/api.c
```diff
@@ -267,125 +339,101 @@ static GSList *scan(struct sr_dev_driver *di, GSList *options)
...
 
-	ftdic = ftdi_new();
-	if (!ftdic) {
-		sr_err("Failed to initialize libftdi.");
-		return NULL;
+	if (conn && conn[0]) {
+		conn_parts = g_strsplit(conn, "/", 2);
+
+		/* USB identifier */
+		if (*conn_parts[0]) {
```
tchebb, line 346:
```suggestion
		if (conn_parts[0][0]) {
```
For consistency

src/hardware/ftdi-la/protocol.c
```diff
@@ -18,88 +19,439 @@
...
+			transfer->status == LIBUSB_TRANSFER_STALL) {
+		sr_err("USB transfer failed: %s.", libusb_error_name(transfer->status));
+		ftdi_la_stop_acquisition(sdi);
+		goto cleanup_transfer;
+	}
+
+	sr_spew("Processing completed transfer of length %d.", transfer->actual_length);
+	send_samples(sdi, transfer->buffer, transfer->actual_length);
+
+	if (!devc->acq_aborted) {
```
tchebb, line 247:
```suggestion
	/* Check this again, since send_samples() may have set it. */
	if (!devc->acq_aborted) {
```

src/hardware/ftdi-la/protocol.c
```diff
@@ -18,88 +19,439 @@
...
+	struct dev_context *devc = sdi->priv;
+	int ret;
+
+	sr_spew("receive_transfer called");
+
+	if (devc->acq_aborted || transfer->status == LIBUSB_TRANSFER_CANCELLED)
+		goto cleanup_transfer;
+
+	if (transfer->status == LIBUSB_TRANSFER_ERROR ||
+			transfer->status == LIBUSB_TRANSFER_NO_DEVICE ||
```
tchebb, line 237: TODO: Test that this works by unplugging the board during a capture.

src/hardware/ftdi-la/protocol.c
```diff
@@ -18,88 +19,439 @@
...
-	/* Get a block of data. */
-	bytes_read = ftdi_read_data(devc->ftdic, devc->data_buf, DATA_BUF_SIZE);
-	if (bytes_read < 0) {
-		sr_err("Failed to read FTDI data (%d): %s.",
-		       bytes_read, ftdi_get_error_string(devc->ftdic));
-		sr_dev_acquisition_stop(sdi);
+	const struct sr_dev_inst *sdi = cb_data;
+	int ret;
+
+	struct timeval tv = { .tv_sec = 0, .tv_usec = 0 };
```
tchebb, line 339: Make const?

src/hardware/ftdi-la/protocol.c
```diff
@@ -18,88 +19,439 @@
...
+
+		libusb_fill_bulk_transfer(devc->transfers[i], usb->devhdl,
+				devc->in_ep_addr, buf, bytes_per_xfer,
+				receive_transfer, (void *)sdi, timeout);
+	}
+
+	return SR_OK;
+}
 
+static int receive_data(int fd, int revents, void *cb_data)
```
tchebb, line 331: Rename to `handle_usb_events()` or something?

tchebb:
Rebased to fix conflicts with master.

tchebb:
Is there anything I can do to make reviewing these changes easier? I've done my best to split the change into individual commits that are independently-reviewable, but I know it's a lot of lines to look through regardless.

I spent a lot of time making and testing these changes, and they bring huge improvements to the FTDI-LA driver, both in terms of correctness (doesn't drop samples anymore) and feature set (higher max sample rate, secondary channels supported). I'd love to see them merged so others can benefit from them.

src/hardware/ftdi-la/api.c
```diff
@@ -39,89 +40,127 @@ static const uint32_t devopts[] = {
...
 static const struct ftdi_chip_desc ft232h_desc = {
 	.vendor = 0x0403,
 	.product = 0x6014,
-	.samplerate_div = 30,
+
+	.multi_iface = TRUE,
+	.num_ifaces = 1,
+
+	/* TODO: This needs validation, since the previous values were weird. */
+	.base_clock = 120000000u,
```
codeottomayer, line 113: Hello,

as this needed validation I have checked an FTDI C232HM-DDHSL-0 devices (which uses a FT232H). I found that the actual sampling rate is about 66 % lower than what is configured in PulseView i.e. an impulse that has duration of 1s (measured on a hardware oscilloscope) is only showing as 666 ms long in PulseView. I have used sample rates of 1 MHz and 100 kHz for this test. I hope this helps.

Best regards,
Otto

tchebb: Thank you so much for the information! I assume those numbers are with these commits applied? Are the timings accurate using the current version of the driver?

I'll try to obtain an FT232H myself so I can dig into what's going on.

codeottomayer: Thank you for your reply Tom!

I have initially tested the above results on the current (download on 11th June) "[Nightly builds](https://sigrok.org/wiki/Downloads)" of PulseView. I have now tested the same using the current (download on 12th June)  "[Nightly builds](https://sigrok.org/wiki/Downloads)" version of "sigrok-cli". I conducted the test using the 64-Bit Windows versions of PulseView and sigrok-cli.

Please do let me know should you require additional info.

Best regards

tchebb: The nightly builds actually don't include this change, since it hasn't been accepted yet. But that's good news, since the error you're seeing is exactly the amount of error I'd expect from the "weird" divisor value in the current driver that my comment mentions.

Here's a version of [sigrok-cli.exe](https://github.com/sigrokproject/libsigrok/files/8897285/sigrok-cli.zip) that includes this change. Do you mind testing it? It ought to fix the sample rate.

While you're going to the trouble of testing it, would you also mind checking if the 15MHz sample rate works? The 12MHz one definitely should, but the 15MHz one might actually run at 7.5MHz (in which case all your waveforms will appear half as long as they should be). If you validate that it does work, though, I can bump up the maximum sample rate for the FT232H in this pull request :)

codeottomayer:
Great Tom! Thank very much you for sending the executable trough. I will get back to you regarding the 15 MHz sample rate within the next 2 to 3 days.

tchebb:
Sorry for the slow response here! It looks like the default Windows build script is using a broken version of libusb. I've tried to switch to a working version (the same on nightly builds use) and rebuilt the exe: try 
[sigrok-cli-20220710.zip](https://github.com/sigrokproject/libsigrok/files/9079646/sigrok-cli-20220710.zip).

codeottomayer:
Hello and apologies for the slow response. I have now tested the version that you have send through. I can confirm that the 12 MHz sample rate works using the C232HM-DDHSL-0 cable i.e. the FTDI FT232 chip.
  
I have also attempted to test the 15 MHz sample rate. However, I did get the following message
 
sr: ftdi-la: Chip does not support sample rate 15000000; adjusted to 12000000.000.

and could therefore not perform the test on the 15 MHz sample rate. 

If you want, you can send through another version and I can check the 15 MHz for you.

tchebb:
> If you want, you can send through another version and I can check the 15 MHz for you.

I appreciate it! My bad for losing the 15MHz change when I fixed the libusb issue. Here's one that should have both: [sigrok-cli-20220730.zip](https://github.com/sigrokproject/libsigrok/files/9227585/sigrok-cli-20220730.zip).

I'm glad to hear that the lower sample rates work well with these changes applied. That confirms my assumption that the existing driver set an incorrect clock divider for FT232H devices.

tchebb:
I've pushed a minor update to this PR that fixes a compilation warning, fixes the build on Windows which uses a very old libusb version, and improves clarity in several places by adding comments or improving code structure.

biot:
Hi Tom, sorry this PR didn't get a proper review sooner. I'd like to help get this merged.

This is an awesome patchset. Not only does it remove support for broken hardware -- a support nightmare -- but it gets rid of libftdi. I never understood why people will use it instead of plain libusb.

i have no comments on any of the patches (!), but the `src/hardware/zeroplus-logic-cube/api.c` patch for `libusb_close()` no longer applies due to changes in mainline. Could you just rebase this?

Thanks for this excellent work!

l29ah:
Pretty sure my "FT232RL"s aren't broken hardware, as i managed to decode plenty of protocols with pulseview's decoders with them (mostly spi and i2c-based). Please don't remove hardware support just because some of the hardware versions aren't up to your standards.

UffeJakobsen:
> Pretty sure my "FT232RL"s aren't broken hardware, as i managed to decode plenty of protocols with pulseview's decoders with them (mostly spi and i2c-based). Please don't remove hardware support just because some of the hardware versions aren't up to your standards.

I agree

Also quoting to the referred errata document "TN_120_FT232R" (https://ftdichip.com/wp-content/uploads/2020/08/TN_120_FT232R-Errata-Technical-Note.pdf):

> The current revision of the FT232R is revision B, released May 2007. At the time of releasing this
> Technical Note there are no known issues with this silicon revision.

tchebb:
@biot thanks so much for the review! I've been busy this last week, but I plan to get this PR back into shape over the next few days. I'd love to see it finally merged!

Gerhard Sittig [reviewed these patches on the mailing list](https://sourceforge.net/p/sigrok/mailman/message/37750418/) in December 2022, and the two of us exchanged some emails at that time. At that time, I pushed a [v2 branch](https://sourceforge.net/p/sigrok/mailman/message/37750418/) to my fork which addressed most (but not all) of his comments and may also fix the merge conflict you pointed out.

Unfortunately, I got busy before I could finish testing the v2 branch (which is why I also didn't update this PR), plus I didn't have a good way to reply to his review on the mailing list since I wasn't subscribed at the time. That's on me, but hopefully this time around I'll have time to get it over the finish line!

@l29ah: Unless your FT232RL is counterfeit, I'm almost certain that it is broken hardware. Every hardware revision of the FT232R has unpredictable bitbang timing. FTDI's errata doc is incorrect that Revision B has no issues, and FTDI's proposed solution—

> Set the clock divisor to 1 (baud rate = 3,000,000) and pad the data field with extra 1’s or 0’s to achieve
the required pulse width for each bit.

—is unworkable for continuous sampling because the FT232R's 12Mbps USB full-speed interface cannot sustain the 24Mbps that would be needed to sample 8 channels at 3MS/s each.

The reason you didn't see issues with (what I assume must have been low-frequency) SPI and I2C is because those protocols both have clock lines, which means the incorrect timing doesn't prevent protocol decode. But just because a certain protocol can still decode doesn't mean that the captures you took weren't objectively incorrect.

And to be clear, I'm not just making a big deal over a sample missing here and there: as you can see in the oscilloscope GIF in [this post](http://blog.bitheap.net/2012/03/ft232r-bitbang-mode-is-broken.html), the timing is grossly incorrect far more often than not, to the point that I think it's fair to say the FT232R is totally unusable as a general-purpose logic analyzer. That's why I decided to drop support in this PR.

I've linked to several more posts and articles detailing the problems with the FT232R's bitbang mode [here on the sigrok wiki](https://sigrok.org/wiki/FTDI-LA#FT232R_Support_Removal), if you want to see more details or evidence.

I did put the removal of support in its own commit, after the bulk of the refactoring, meaning it should cleanly revert for those who know what they're doing and have a use case where sample timings don't matter. But for those who don't know the risks, I think dropping support by default will cause far less confusion and frustration than keeping it would.

UffeJakobsen:
> 
> Unless your FT232RLs are counterfeit, I'm almost certain that they are broken hardware. Every hardware revision of the FT232R has unpredictable bitbang timing. FTDI's errata doc is incorrect that Revision B has no issues, and FTDI's proposed solution—
> 
I do know that some FT232RL's has broken bitbang feature. But I do also have units where bitbanging - that would otherwise fail using another unit - just works

abraxa:
@tchebb It appears the patch set has conflicts and won't apply cleanly. Can you please fix and rebase?

That said, I'd like to keep FR232RL support if it doesn't make the code overly complicated. The users should, however, be made aware of the issue somehow. Maybe the wiki page is the best place to do so. The driver description could use amending, in my opinon, since "FTDI LA" is honestly very nondescript. It could be a good idea to change it to "FTDI FT232/2232 logic analyzer (see device page on the wiki for important notes)" or similar.

biot:
> That said, I'd like to keep FR232RL support if it doesn't make the code overly complicated. The users should, however, be made aware of the issue somehow.

The problem with this is that the current broken situation just goes on as before, and people will inevitably run into broken LA sessions which they don't know is due to these FT232R, not sigrok. So I think we need people that still want this to jump through some kind of hoop first.

I'm thinking along the lines of the cmdline option some tools have where you need to type `--yes-i-know-this-is-a-bad-idea` for some ill-advised feature. The only thing I can think of is having to specify `forced_model`  as `ft232r`, with the driver refusing to use it otherwise. Does that sound like a good compromise?

biot:
Errm, the option is actually called `SR_CONF_FORCE_DETECT`, not force_model :-)

tchebb:
I haven't forgotten about this PR; still planning to rebase and retest once I get the time.

Since the FT232R removal is controversial, I'll drop that commit from this PR and move it into a separate one. Gating it behind some kind of "I know what I'm doing" flag is fine by me!

PR #147, opened 2021-06-23
SR_CONF_xxx: Clarify rationale behind unusual key

karlp:
As reported on IRC, meaning of SR_CONF_SAMPLE_INTERVAL is unclear.  We already have doxygen comments, we can use them, and rather than simply repeating the field name, provide some context.

### DIFF
--- a/include/libsigrok/libsigrok.h
+++ b/include/libsigrok/libsigrok.h
@@ -836,7 +836,11 @@ enum sr_configkey {
 	/** Trigger matches. */
 	SR_CONF_TRIGGER_MATCH,
 
-	/** The device supports setting its sample interval, in ms. */
+	/**
+	 * The device supports setting its sample interval, in ms.
+	 * This is only used for devices that only support <1Hz sample rates.
+	 * @sa SR_CONF_SAMPLERATE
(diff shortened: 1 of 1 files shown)


include/libsigrok/libsigrok.h
```diff
@@ -836,7 +836,11 @@ enum sr_configkey {
 	/** Trigger matches. */
 	SR_CONF_TRIGGER_MATCH,
 
-	/** The device supports setting its sample interval, in ms. */
+	/**
+	 * The device supports setting its sample interval, in ms.
+	 * This is only used for devices that only support <1Hz sample rates.
+	 * @sa SR_CONF_SAMPLERATE
```
knarfS, line 842: Could you also add a cross reference from SR_CONF_SAMPLERATE  to this config key?

knarfS (changes requested):

include/libsigrok/libsigrok.h
karlp: not really?  I don't even have this code locally. consider this patch public domain, I wsa just trying to help it along.

PR #149, opened 2021-06-24
Add support for another open-source logic analyzer project (open-logic-bit)

ultraembedded:
Hi,

I have added support for another open-source logic analyzer project (open-logic-bit) to libsigrok.
The aim of this project is to support alternate gateware (i.e. FPGA bitstreams) for various FPGA dev boards and also other LA's which are not currently supported by libsigrok.

The gateware (Verilog RTL) project is located here:
https://github.com/ultraembedded/openlogicbit

The first target that I have this working on is the Digilent Digital Discovery unit which libsigrok does not currently support with the vendor-supplied gateware/protocol. This unit is a 24 channel, 100MHz + LA with 256MB of DDR3 memory.
With my replacement gateware + libsigrok hardware driver, this unit is working in sigrok/pulseview with good stability (it's really quite usable!), albeit with quite basic LA features.
With this board + open-logic-bit gateware, up to 256MB of RLE encoded data can be captured at up to 100MHz, and can be downloaded from the device at between 20-30MB/s.

Currently, my project is aimed at FPGA boards / LAs with some kind of large capture memory (SDRAM, DDR), and with the FTDI FT232H/FT2232H USB2.0 interface.
As time goes on, I will add new trigger modes (the current implementation is quite basic) and new supported boards and host interfaces (such as the other FTDI parts including the USB3.0 device).

I am aware that there are other open-source logic analyzers such as the Open Bench Logic Sniffer / SUMP stuff - but these tend to rely on either small built-in FPGA memories (i.e. limited capture buffer depth), or have very slow host interfaces (UART).

Please let me know if I need to fix anything with this PR - I did try to follow the libsigrok coding guidelines, and I've done a good amount of testing.


wsakernel:
Looks really interesting. I'll try to have a look at the code in the next weeks. I can't really promise when, but I'll try.

biot:
This looks like an amazing project. I'm sorry this didn't get reviewed before. Do you still want this driver merged? The open-logic-bit project doesn't appear to have been very active lately either, and I see no other boards got added.

If this is still something you want to continue, I'd love to review the code (it's good, just a few minor coding style fixes). I also have some suggestions, about the project rather than the code:

- The protocol really needs some kind of "ID" command, where the gateware responds with details about which board it's running on, particularly important since USB vid/pid are so generic in the Digilent board.
- I would also add a more general "get configuration" command. Look at the OLS metadata struct for what all could go in there (that struct was done a bit ad-hoc, could be better).

PR #157, opened 2021-08-25
input/vcd: Ignore string value change entries

antonblanchard:
Some tools (eg nmigen) output string value change entries. Instead of
failing, just skip over them.

### DIFF
--- a/src/input/vcd.c
+++ b/src/input/vcd.c
@@ -1518,7 +1518,7 @@ static int parse_textline(const struct sr_input *in, char *lines)
 	char **words;
 	size_t word_count, word_idx;
 	char *curr_word, *next_word, curr_first;
-	gboolean is_timestamp, is_section, is_real, is_multibit, is_singlebit;
+	gboolean is_timestamp, is_section, is_real, is_multibit, is_singlebit, is_string;
 	uint64_t timestamp;
 	char *identifier, *endptr;
 	size_t count;
@@ -1725,6 +1725,7 @@ static int parse_textline(const struct sr_input *in, char *lines)
 		is_singlebit |= curr_first == 'l' || curr_first == 'h';
 		is_singlebit |= curr_first == 'x' || curr_first == 'z';
 		is_singlebit |= curr_first == 'u' || curr_first == '-';
+		is_string = curr_first == 's';
 		if (is_real) {
 			char *real_text;
 			float real_val;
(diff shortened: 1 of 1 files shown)


wsakernel:
Looks good to me. Could you kindly provide an example file to test the change?

antonblanchard:
Sure! Where should this live? In `sigrok-test`? Did you want me to submit a separate PR?

```
$comment Generated by nMigen $end
$date 2021-09-04 08:55:28.031092 $end
$timescale 100 ps $end
$scope module top $end
$var wire 1 ! clk $end
$var wire 1 " rst $end
$var wire 1 # val_out $end
$var string 1 $ state $end
$upscope $end
$enddefinitions $end
#0
$dumpvars
0!
0"
0#
sIDLE/0 $
$end
#50
1#
sBUSY/1 $
1!
#100
0!
#150
0#
sIDLE/0 $
1!
#200
0!
#250
1#
sBUSY/1 $
1!
#300
```

wsakernel:
I think here is fine. It is actually a good idea to put test data for input/output-filters into sigrok-test as well. But we would first need infrastructure to make use of the data. So, for now, what you pasted above seems fine to me, thanks. I will test it later.

wsakernel (approved):
From my side, everything is fine this way. Let's see what Gerhard thinks.

gsigh:
(This was discussed in IRC on 2021-09 but didn't make it here.)

The report is valuable to raise awareness of that VCD feature. But the suggested implementation is not acceptable to sigrok mainline. It introduces unreliability to an existing implementation which tries to be as robust as possible. Silently(!) dropping input data is a blocker. Especially when the condition which results in dropped data is so weak. I do understand that the feature is rare and may be underspecified, so that the approach taken here might have been tempting at the time of creation of that commit.

For illustration: The suggested implementation would "accept" input like this

    ... 0! 1" something invalid 1! 0" ...
    ... 0! 1" some bug says hello 1! 0" ...

and would drop input data _without_ telling the user. This is unexpected, and should be avoided.

I got a more conservative approach in my local queue, but it's essential to learn what exactly string data in VCD input looks like. Is there some spec or some "kind of official" paper which gives us more certainty about what to expect, and what to reject? The fact that some generator emits these doesn't mean much, could have been some local extension. The gtkwave code which reads the input keeps the support under "not strict VCD" which doesn't give much confidence either. Ideally there'd be a better source of information than some random implementation.

gsigh:
I'd suggest to put the sample file in the dumps repository. Even if it's not used in the test suite, it's useful to have around and available to users, too. After all it demonstrates an exceptional condition and rare feature that is used "in the wild".

gsigh:
This is what I came up with in September, and need more feedback on so that we can create something which is acceptable for sigrok mainline.  https://repo.or.cz/libsigrok/gsi.git/shortlog/refs/heads/wip/vcd-string-type-v1

gsigh:
Commit c1310f7deba7 adds logic to gracefully skip string values in VCD input files, yet keeps robustly rejecting invalid input for users' awareness. This is the best that I could come up with in the absence of an official spec or any public information on what "strings" might mean in the context of VCD files.

PR #158, opened 2021-09-06
output/csv: fix dedup logic

kelvie:
This fixes the dedup logic in the csv output to:
- dedup across samples (much like the vcd output module)
- not print two values for every group of samples

`previous_sample` was being freed every receive, and so it was
impossible to dedup across multiple receives, which would cause many
rows even when the level hasn't changed when using high sample rates.

There was also a range check for `i` that makes it so that two values
will always skip the dedup check, again spamming rows even when the
level hasn't changed.

### DIFF
--- a/src/output/csv.c
+++ b/src/output/csv.c
@@ -458,6 +458,8 @@ static void dump_saved_values(struct context *ctx, GString **out)
 			    &ctx->logic_samples[i * ctx->num_logic_channels];
 
 
+			ctx->out_sample_count++;
+
 			if (ctx->dedup) {
 				if (!memcmp(logic_sample, ctx->previous_sample,
 					    ctx->num_logic_channels) &&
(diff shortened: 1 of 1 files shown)


src/output/csv.c
```diff
@@ -513,12 +513,10 @@ static void dump_saved_values(struct context *ctx, GString **out)
 	}
 
 	/* Discard all of the working space. */
-	g_free(ctx->previous_sample);
```
kelvie, line 516: This is free'd in the cleanup function so I don't think there's a memory leak.

src/output/csv.c
```diff
@@ -476,7 +478,7 @@ static void dump_saved_values(struct context *ctx, GString **out)
 			if (ctx->time && !ctx->sample_rate) {
 				g_string_append_printf(*out, "0%s", ctx->value);
 			} else if (ctx->time) {
-				sample_time_dbl = ctx->out_sample_count++;
+				sample_time_dbl = ctx->out_sample_count - 1;
```
biot, line 481: You're correct that the sample count needs to be incremented earlier, but why does the `- 1` happen at all?

biot (changes requested):
I'm sorry your PR didn't get reviewed sooner, I hope you're still interested in fixing the CSV output module.

However your fixes need some work.

kelvie:
Hello, unfortunately that thing happened where I've forgotten how I encountered this fix after 3 years . I'll close this, but yeah feel free to incorporate these changes if they're helpful (I don't need attribution).

PR #159, opened 2021-09-26
Python bindings: fix setting of rational values

silbe:
Setting of rational voltage values like `vdiv` (Volts/division) is broken in the Python bindings because the GVariant type created by the Python bindings is rejected by libsigrok (tested with `rigol-ds`). This patch fixes the mismatch by using the same code as the C++ bindings.

In addition rational period values like `timebase` were not handled at all by the Python bindings. This is fixed by handling them the same way as rational voltage values.


rsaxvc:
@silbe , I'm not an expert in glib, or libsigrok, but this seems reasonable and cleaner  to me.

PR #131, opened 2021-04-05
Kingst la2016 driver fixup

Planet911:
The sigrok driver for the Kingst LA2016 is currently broken for many use cases, it doesn't work at all for me in PulseView.
This pr fixes a number of driver issues, mainly the fpga spi register addressing, and also adds support for older LA2016 devices.
It also adds support for LA1016 devices (old and new revisions) which have the same hardware as LA2016 devices but use different fpga bitstreams and are limited to 100MHz sampling.
This has all been tested on LA2016 and LA1016 devices, on linux host but I don't envisage any portability issues.

PR #161, opened 2021-10-02
kingst-la2016: Fix USB issues on Win build

Planet911:
My [previous pull request](https://github.com/sigrokproject/libsigrok/pull/131) for the Kingst LA2016 fixed some driver issues and got the device working in nightly builds for Linux. However, the driver did not work in Windows due to a USB transfer problem which this PR fixes.

The issue was related to the special build of libusb which sigrok uses on Windows, which enables WinUSB RAW_IO policy. This requires the USB transfer buffer size to be multiple of endpoint max packet size.

This PR is tiny and (trust me!) it will get the Windows nightly build working. Tested with LA2016 and LA1016 on Linux and Windows. I would appreciate it if this could be pulled relatively quickly.

### DIFF
--- a/src/hardware/kingst-la2016/api.c
+++ b/src/hardware/kingst-la2016/api.c
@@ -641,8 +641,11 @@ static void LIBUSB_CALL receive_transfer(struct libusb_transfer *transfer)
 	devc->n_bytes_to_read -= transfer->actual_length;
 	if (devc->n_bytes_to_read) {
 		uint32_t to_read = devc->n_bytes_to_read;
-		if (to_read > LA2016_BULK_MAX)
-			to_read = LA2016_BULK_MAX;
+		/* determine read size for the next usb transfer */
+		if (to_read >= LA2016_USB_BUFSZ)
+			to_read = LA2016_USB_BUFSZ;
+		else /* last transfer, make read size some multiple of LA2016_EP6_PKTSZ */
+			to_read = (to_read + (LA2016_EP6_PKTSZ-1)) & ~(LA2016_EP6_PKTSZ-1);
 		libusb_fill_bulk_transfer(
 			transfer, usb->devhdl,
 			0x86, transfer->buffer, to_read,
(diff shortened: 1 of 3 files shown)


abraxa:
Thanks for diving into this, I know how much of a pain Windows builds unfortunately are! It's merged as e847645b7be832c1d62d1ad2312e3a3d4de467ce and nightly builds should be updated in about 9 hours from now.

src/hardware/kingst-la2016/api.c
```diff
@@ -641,8 +641,11 @@ static void LIBUSB_CALL receive_transfer(struct libusb_transfer *transfer)
 	devc->n_bytes_to_read -= transfer->actual_length;
 	if (devc->n_bytes_to_read) {
 		uint32_t to_read = devc->n_bytes_to_read;
-		if (to_read > LA2016_BULK_MAX)
-			to_read = LA2016_BULK_MAX;
+		/* determine read size for the next usb transfer */
+		if (to_read >= LA2016_USB_BUFSZ)
+			to_read = LA2016_USB_BUFSZ;
```
wsakernel, line 646: '>' would be enough

src/hardware/kingst-la2016/protocol.c
```diff
@@ -723,9 +723,11 @@ SR_PRIV int la2016_start_retrieval(const struct sr_dev_inst *sdi, libusb_transfe
...
 
 	to_read = devc->n_bytes_to_read;
-	if (to_read > LA2016_BULK_MAX)
-		to_read = LA2016_BULK_MAX;
-
+	/* choose a buffer size for all of the usb transfers */
+	if (to_read >= LA2016_USB_BUFSZ)
+		to_read = LA2016_USB_BUFSZ; /* multiple transfers */
+	else /* one transfer, make buffer size some multiple of LA2016_EP6_PKTSZ */
+		to_read = (to_read + (LA2016_EP6_PKTSZ-1)) & ~(LA2016_EP6_PKTSZ-1);
```
wsakernel, line 730: It probably makes sense to to turn the new code into an inline function? So we make sure that both invocations are always consistent.

wsakernel (commented):
I know this request has already been applied. But my comments could be added on top of this if we can agree on them.

src/hardware/kingst-la2016/protocol.c
Planet911: Your right, I should have done it that way. There are a few things that can be improved in the LA2016 driver, but at least it is functional for now. I would like to get the open mcu firmware I have for LA2016 accepted into sigrok, then look at tidy ups.

PR #162, opened 2021-10-24
fixing memory leak in lecroy-xstream

LucienMP:
The character data used from the sr_scpi_get_string routine performs a strtok and is passed into several pointers, and incorrectly g_free'd resulting in invalid pointer data being used, and potential memory leaks/crashes.


wsakernel (changes requested):

biot:
This patch no longer cleanly applies on current master, and somebody fixed half of the problem.

I've submitted a patch to finally fix this memory leak, inspired by @LucienMP's patch and @wsakernel's suggestion. Thanks!

PR #165, opened 2021-12-03
zeroplus-logic-cube: Check vid and pid before string descriptors

kfazi:
This PR moves the checking of VID and PID to the beginning of the scan function.

On my machine, I have an AURA LED Controller.
It causes SEGV in libusb when reading its string descriptors.
Moving the VID and PID checking allows filtering out this controller and workaround the problem.

As a side-effect, device scanning should speed up a bit.

This together with https://github.com/sigrokproject/libsigrok/pull/166 should fix https://sigrok.org/bugzilla/show_bug.cgi?id=1115

### DIFF
--- a/src/hardware/zeroplus-logic-cube/api.c
+++ b/src/hardware/zeroplus-logic-cube/api.c
@@ -184,6 +184,17 @@ static GSList *scan(struct sr_dev_driver *di, GSList *options)
 	for (i = 0; devlist[i]; i++) {
 		libusb_get_device_descriptor(devlist[i], &des);
 
+		prof = NULL;
+		for (j = 0; j < zeroplus_models[j].vid; j++) {
+			if (des.idVendor == zeroplus_models[j].vid &&
+				des.idProduct == zeroplus_models[j].pid) {
+				prof = &zeroplus_models[j];
+			}
+		}
+
+		if (!prof)
+			continue;
+
 		if ((ret = libusb_open(devlist[i], &hdl)) < 0)
 			continue;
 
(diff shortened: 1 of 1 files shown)


wsakernel (approved):
Heh, nice one. Selling the workaround as a speed-up. Works for me :)
However, I still wonder if the proper thing to do would be fixing libusb?

kfazi:
@wsakernel yes, it would be best to fix libusb, but I couldn't reproduce this bug in isolation. A simple application that reads string descriptors of all devices works as expected.

Something in libsigrok (or pulseview) must be changing libusb internal state, that it freaks out on Aura Controller.

wsakernel:
Kudos that you already created an extra application. Pity it didn't reveal something yet.
Have you tried using sigrok-cli? I am quite sure it will neither work. I'd be really surprised if pulseview is the culprit.

kfazi:
@wsakernel What can I do to have this PR merged?

fenugrec:
I'm totally in favour of validating VID+PID before trying to open a device. Looks like a fairly trivial changeset. +1 from me

fenugrec:
However, the original code is strangely written :

```
		for (j = 0; j < zeroplus_models[j].vid; j++) {
			if (des.idVendor == zeroplus_models[j].vid &&
				des.idProduct == zeroplus_models[j].pid) {
				prof = &zeroplus_models[j];
			}
		}
```

What ? `  j < zeroplus_models[j].vid ` ? Sure, it works because the last array item has `.vid` = 0, and by then `j` has incremented a few times already. But IMO could be written simpler,

`for (j = 0; zeroplus_models[j].vid; j++) {`

fenugrec:
Can the interested parties test this branch , that addresses these problems , but goes a bit beyond in cleaning up surrounding code :
https://repo.or.cz/libsigrok/gsi.git/shortlog/refs/heads/wip/chronovu-cleanup

(if you're like me and always forget how to fetch just a branch, here you go :
```
 git remote add gsi https://repo.or.cz/libsigrok/gsi.git
 git fetch gsi wip/chronovu-cleanup
 git checkout FETCH_HEAD
```

biot:
A fix for this was merged in https://github.com/sigrokproject/libsigrok/commit/fd86380b48f1abff31f523e33c7c12280be1c1f7

PR #166, opened 2021-12-03
chronovu-la: Add checking of VendorId

kfazi:
On my machine, I have an AURA LED Controller.
It causes SEGV in libusb when reading its string descriptors.

Checking the VID during the scan filters out this controller and workarounds the problem. I chose to filter only VID as I don't own any device of this type and I couldn't find if there are multiple PIDs for these devices.
The VID is taken from https://sigrok.org/wiki/ChronoVu_LA8/Info and https://sigrok.org/wiki/ChronoVu_LA16/Info.

As a side-effect, device scanning should speed up a bit.

This together with https://github.com/sigrokproject/libsigrok/pull/165 should fix https://sigrok.org/bugzilla/show_bug.cgi?id=1115

### DIFF
--- a/src/hardware/chronovu-la/api.c
+++ b/src/hardware/chronovu-la/api.c
@@ -20,6 +20,8 @@
 #include <config.h>
 #include "protocol.h"
 
+#define CHRONOVU_VENDOR (0x0403)
+
 static const uint32_t scanopts[] = {
 	SR_CONF_CONN,
 };
@@ -168,6 +170,9 @@ static GSList *scan(struct sr_dev_driver *di, GSList *options)
 		}
 
 		libusb_get_device_descriptor(devlist[i], &des);
+
+		if (des.idVendor != CHRONOVU_VENDOR)
+			continue;
 
 		if ((ret = libusb_open(devlist[i], &hdl)) < 0)
 			continue;


wsakernel:
The main issue I have with this PR is the workaround nature of it. We will run into bug 1115 again as soon as another new driver reads out the strings. So, fixing the root cause is key. We could apply this PR to buy us some time, but then there should be a comment around the newly added code which points to bug 1115. And bug 1115 should not be closed until we fixed the root cause. That's my 2 cents about it. Opinions?

kfazi:
It seems that this LED Controller I have on my motherboard is out of spec, and there is not much we can do to fix it.

The libusb folks say they are not interested in supporting broken devices. See the issue https://github.com/libusb/libusb/issues/725.

Limiting drivers to query devices with a specific VID/PID is a way to avoid touching unknown devices.

kfazi:
@wsakernel What can I do to have this PR merged?

abraxa:
Merged as https://sigrok.org/gitaction?p=libsigrok.git;a=commit;h=64414853067e727f99c7d41851ac635d4342f17b, thank you!

I added references to #1115 and this PR as a comment.

PR #168, opened 2022-01-05
dslogic: Optimize deinterleave_buffer

christianeisendle:
deinterleave_buffer is used to convert the received samples from DSLogic device (which is a sequence of 64-bit words in round-robin for each channel, to the format sigrok is expecting (sequence of words containing the value for given channel as bitmask).

The existing implementation looped over every single bit, making it inefficient assuming that a typical trace contains little signal changes compared to the sample rate.

This change improves the deainterleave function by forward copying 64 words assuming there is no change. Only if there is a change, it searches the individual channels for bit-wise changes.


christianeisendle:
Closing this one since it got superseded by #170

PR #169, opened 2022-01-11
kingst-la2016: fix configuration bytes for huge captures

Planet911:
The sample counts are 34bits rather than 32bits.
Tested successfully on LA2016 and LA1016.


gsigh:
See commit b1a17c1a2026 which uses the information that you provided and implements the logic in more reliable and more readable and thus more maintainable ways. I gave you a Reported-By: line for your submission. Thank you!

PR #170, opened 2022-01-13
Support for dreamsourcelab DSLogic U3Pro16

christianeisendle:
This PR adds support for DSLogic U3Pro16 to libsigrok.
It is a logic analyzer with up to 1GHz sample rate 16 channels, buffered and streaming mode support and USB3 interface.

The support has been added to existing dreamsourcelab-dslogic driver with some runtime diversity since the U3Pro16 has a new API which is not fully backward-compatible with the legacy dreamsource logic devices.

In addition, the generic part of the driver got some performance updates in the deinterleaving function and a dedicated worker thread to process/deinterlave samples while processing already the next USB transfer.

### DIFF
--- a/contrib/60-libsigrok.rules
+++ b/contrib/60-libsigrok.rules
@@ -110,6 +110,8 @@ ATTRS{idVendor}=="2a0e", ATTRS{idProduct}=="0002", ENV{ID_SIGROK}="1"
 ATTRS{idVendor}=="2a0e", ATTRS{idProduct}=="0020", ENV{ID_SIGROK}="1"
 # DreamSourceLab DSLogic Basic
 ATTRS{idVendor}=="2a0e", ATTRS{idProduct}=="0021", ENV{ID_SIGROK}="1"
+# DreamSourceLab DSLogic U3Pro16
+ATTRS{idVendor}=="2a0e", ATTRS{idProduct}=="002a", ENV{ID_SIGROK}="1"
 
 # GW-Instek GDM-9061 (USBTMC mode)
 ATTRS{idVendor}=="2184", ATTRS{idProduct}=="0059", ENV{ID_SIGROK}="1"
--- a/src/hardware/dreamsourcelab-dslogic/api.c
+++ b/src/hardware/dreamsourcelab-dslogic/api.c
@@ -51,7 +51,7 @@ static const struct dslogic_profile supported_device[] = {
 	/* DreamSourceLab DSLogic U3Pro16 */
 	{ 0x2a0e, 0x002a, "DreamSourceLab", "DSLogic U3Pro16", NULL,
 		"dreamsourcelab-dslogic-basic-fx2.fw",
-		DSLOGIC_CAPS_ADF4360, "DreamSourceLab", "DSLogic", 
+		DSLOGIC_CAPS_ADF4360 | DSLOGIC_CAPS_USB30, "DreamSourceLab", "DSLogic", 
 		2ULL * 1024ULL * 1024ULL * 1024ULL, 40,
 		1024, SR_MHZ(500), SR_MHZ(500), SR_GHZ(1), 5},
 
--- a/src/hardware/dreamsourcelab-dslogic/api.c
+++ b/src/hardware/dreamsourcelab-dslogic/api.c
@@ -26,23 +26,34 @@ static const struct dslogic_profile supported_device[] = {
 	/* DreamSourceLab DSLogic */
 	{ 0x2a0e, 0x0001, "DreamSourceLab", "DSLogic", NULL,
 		"dreamsourcelab-dslogic-fx2.fw",
-		0, "DreamSourceLab", "DSLogic", 256 * 1024 * 1024},
+		0, "DreamSourceLab", "DSLogic", 256 * 1024 * 1024, 100,
+		512, DS_API_V1, SR_MHZ(100), SR_MHZ(200), SR_MHZ(400), 1},
 	/* DreamSourceLab DSCope */
 	{ 0x2a0e, 0x0002, "DreamSourceLab", "DSCope", NULL,
 		"dreamsourcelab-dscope-fx2.fw",
-		0, "DreamSourceLab", "DSCope", 256 * 1024 * 1024},
+		0, "DreamSourceLab", "DSCope", 256 * 1024 * 1024, 100,
+		512, DS_API_V1, SR_MHZ(100), SR_MHZ(200), SR_MHZ(400), 1},
 	/* DreamSourceLab DSLogic Pro */
 	{ 0x2a0e, 0x0003, "DreamSourceLab", "DSLogic Pro", NULL,
 		"dreamsourcelab-dslogic-pro-fx2.fw",
-		0, "DreamSourceLab", "DSLogic", 256 * 1024 * 1024},
+		0, "DreamSourceLab", "DSLogic", 256 * 1024 * 1024, 100,
+		512, DS_API_V1, SR_MHZ(100), SR_MHZ(200), SR_MHZ(400), 1},
 	/* DreamSourceLab DSLogic Plus */
 	{ 0x2a0e, 0x0020, "DreamSourceLab", "DSLogic Plus", NULL,
 		"dreamsourcelab-dslogic-plus-fx2.fw",
-		0, "DreamSourceLab", "DSLogic", 256 * 1024 * 1024},
+		0, "DreamSourceLab", "DSLogic", 256 * 1024 * 1024, 100,
+		512, DS_API_V1, SR_MHZ(100), SR_MHZ(200), SR_MHZ(400), 1},
 	/* DreamSourceLab DSLogic Basic */
 	{ 0x2a0e, 0x0021, "DreamSourceLab", "DSLogic Basic", NULL,
 		"dreamsourcelab-dslogic-basic-fx2.fw",
-		0, "DreamSourceLab", "DSLogic", 256 * 1024},
+		0, "DreamSourceLab", "DSLogic", 256 * 1024, 100,
+		512, DS_API_V1, SR_MHZ(100), SR_MHZ(200), SR_MHZ(400), 1},
+	/* DreamSourceLab DSLogic U3Pro16 */
+	{ 0x2a0e, 0x002a, "DreamSourceLab", "DSLogic U3Pro16", NULL,
+		"dreamsourcelab-dslogic-basic-fx2.fw",
+		DSLOGIC_CAPS_ADF4360, "DreamSourceLab", "DSLogic", 
+		2ULL * 1024ULL * 1024ULL * 1024ULL, 40,
+		1024, DS_API_V2, SR_MHZ(500), SR_MHZ(500), SR_GHZ(1), 5},
 
 	ALL_ZERO
 };
@@ -102,6 +113,27 @@ static const uint64_t samplerates[] = {
 	SR_MHZ(100),
 	SR_MHZ(200),
 	SR_MHZ(400),
+};
+
+static const uint64_t samplerates_U3Pro16[] = {
+	SR_KHZ(10),
+	SR_KHZ(20),
+	SR_KHZ(50),
+	SR_KHZ(100),
+	SR_KHZ(200),
+	SR_KHZ(500),
+	SR_MHZ(1),
+	SR_MHZ(2),
+	SR_MHZ(5),
+	SR_MHZ(10),
+	SR_MHZ(20),
+	SR_MHZ(25),
+	SR_MHZ(50),
+	SR_MHZ(100),
+	SR_MHZ(125),
+	SR_MHZ(250),
+	SR_MHZ(500),
+	SR_MHZ(1000),
 };
 
 static gboolean is_plausible(const struct libusb_device_descriptor *des)
@@ -255,9 +287,18 @@ static GSList *scan(struct sr_dev_driver *di, GSList *options)
 		sdi->priv = devc;
 		devices = g_slist_append(devices, sdi);
 
-		devc->samplerates = samplerates;
-		devc->num_samplerates = ARRAY_SIZE(samplerates);
-		has_firmware = usb_match_manuf_prod(devlist[i], "DreamSourceLab", "USB-based Instrument");
+		if (strcmp("DSLogic U3Pro16", devc->profile->model) == 0) {
+			devc->samplerates = samplerates_U3Pro16;
+			devc->num_samplerates = ARRAY_SIZE(samplerates_U3Pro16);
+		} else {
+			devc->samplerates = samplerates;
+			devc->num_samplerates = ARRAY_SIZE(samplerates);
+
+		}
+		if (devc->profile->api_version == DS_API_V1)
+			has_firmware = usb_match_manuf_prod(devlist[i], "DreamSourceLab", "USB-based Instrument");
+		else
+			has_firmware = usb_match_manuf_prod(devlist[i], "DreamSourceLab", "USB-based DSL Instrument v2");
 
 		if (has_firmware) {
 			/* Already has the firmware, so fix the new address. */
--- a/src/hardware/dreamsourcelab-dslogic/api.c
+++ b/src/hardware/dreamsourcelab-dslogic/api.c
@@ -477,6 +477,9 @@ static int config_get(uint32_t key, GVariant **data,
 	case SR_CONF_EXTERNAL_CLOCK:
 		*data = g_variant_new_boolean(devc->external_clock);
 		break;
+	case SR_CONF_CONTINUOUS:
+		*data = g_variant_new_boolean(devc->continuous_mode);
+		break;
 	case SR_CONF_CLOCK_EDGE:
 		idx = devc->clock_edge;
 		if (idx >= (int)ARRAY_SIZE(signal_edges))
@@ -529,6 +532,9 @@ static int config_set(uint32_t key, GVariant *data,
 		break;
 	case SR_CONF_EXTERNAL_CLOCK:
 		devc->external_clock = g_variant_get_boolean(data);
+		break;
+	case SR_CONF_CONTINUOUS:
+		devc->continuous_mode = g_variant_get_boolean(data);
 		break;
 	case SR_CONF_CLOCK_EDGE:
 		if ((idx = std_str_idx(data, ARRAY_AND_SIZE(signal_edges))) < 0)
--- a/src/hardware/dreamsourcelab-dslogic/api.c
+++ b/src/hardware/dreamsourcelab-dslogic/api.c
@@ -27,33 +27,33 @@ static const struct dslogic_profile supported_device[] = {
 	{ 0x2a0e, 0x0001, "DreamSourceLab", "DSLogic", NULL,
 		"dreamsourcelab-dslogic-fx2.fw",
 		0, "DreamSourceLab", "DSLogic", 256 * 1024 * 1024, 100,
-		512, DS_API_V1, SR_MHZ(100), SR_MHZ(200), SR_MHZ(400), 1},
+		512, SR_MHZ(100), SR_MHZ(200), SR_MHZ(400), 1},
 	/* DreamSourceLab DSCope */
 	{ 0x2a0e, 0x0002, "DreamSourceLab", "DSCope", NULL,
 		"dreamsourcelab-dscope-fx2.fw",
 		0, "DreamSourceLab", "DSCope", 256 * 1024 * 1024, 100,
-		512, DS_API_V1, SR_MHZ(100), SR_MHZ(200), SR_MHZ(400), 1},
+		512, SR_MHZ(100), SR_MHZ(200), SR_MHZ(400), 1},
 	/* DreamSourceLab DSLogic Pro */
 	{ 0x2a0e, 0x0003, "DreamSourceLab", "DSLogic Pro", NULL,
 		"dreamsourcelab-dslogic-pro-fx2.fw",
 		0, "DreamSourceLab", "DSLogic", 256 * 1024 * 1024, 100,
-		512, DS_API_V1, SR_MHZ(100), SR_MHZ(200), SR_MHZ(400), 1},
+		512, SR_MHZ(100), SR_MHZ(200), SR_MHZ(400), 1},
 	/* DreamSourceLab DSLogic Plus */
 	{ 0x2a0e, 0x0020, "DreamSourceLab", "DSLogic Plus", NULL,
 		"dreamsourcelab-dslogic-plus-fx2.fw",
 		0, "DreamSourceLab", "DSLogic", 256 * 1024 * 1024, 100,
-		512, DS_API_V1, SR_MHZ(100), SR_MHZ(200), SR_MHZ(400), 1},
+		512, SR_MHZ(100), SR_MHZ(200), SR_MHZ(400), 1},
 	/* DreamSourceLab DSLogic Basic */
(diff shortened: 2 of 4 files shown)


Electro707:
Thanks for this PR! I was trying to port the device a while by attempting to copy the codebase from DSView, but that didn't work out to well. 

I am currently trying this fork on my U3Pro16 (unfortunately I don't have any older devices to test for legacy), and here are a couple of things I noticed:
- There is a sample rate limit per number of channels enabled. I do not see something implemented in here, so I am not sure if what the logic analyzer is returning when sampling at 500Mhz for example with all channels enabled are correct. I am currently only testing it with one channel, and that at least seems right.
- At 1Ghz on all channels enabled, I get some weird response on the other channels
  - Normal/Expected: image
  - At 1Ghz: image
- If an acquisition is started but stopped with the Stop button on Pulseview, all acquisition data is lost.
- When acquiring from the logic analyzer, PulseView does not update with the current collected data. This is a minor point, but it does relate to the previous point.

christianeisendle:
>There is a sample rate limit per number of channels enabled. I do not see something implemented in here, so I am not sure if what the logic analyzer is returning when sampling at 500Mhz for example with all channels enabled are correct. I am currently only testing it with one channel, and that at least seems right.

Right, the U3Pro16 has some HW limits which is a combination of enabled channels and configured sample rate. That's also the reason why in DSView some combinations don't work. pulseview on the other hand is device agnostic, so it has no idea about any combinatorial device constraints. On startup it fetches the supported sample rate from libsigrok and the number of channels, but completely independent from each other. I could add some checks in the driver the disallow some combinations and return an error to the upper application. I guess in sigrok-cli this error is shown but in pulseview it would just get logged on the console/logger but not in the GUI. So at the end, pulseview would need an update as well but since pulseview is device agnostic it would mean that such change has a generic impact on all devices. But let me look into this to see which combinations are allowed.

>At 1Ghz on all channels enabled, I get some weird response on the other channels 
This is exactly the problem I mentioned above. At 1GHz you can't use all channels (DSView for instance limits it to 8 channels in buffered mode and i guess 3 channels in streaming mode). 

> If an acquisition is started but stopped with the Stop button on Pulseview, all acquisition data is lost.
Indeed, in this case I stop the ongoing acquisition but don't start the USB transfer to fetch the data. I will change this.

> When acquiring from the logic analyzer, PulseView does not update with the current collected data. This is a minor point, but it does relate to the previous point.
That's a legacy limitation along with another limitation from pulseview. U3Pro16 supports three modes:
- Buffered mode
- Streaming mode
- Test mode
If you look into DSView you can switch between these modes.
Some details on it
- Buffered mode: In this mode the channels are captured and stored in internal RAM of U3Pro. This is done until internal memory is full and only then a USB transfer is started to transfer the data from U3Pro RAM to the PC. In this case you see the data updated only after the capture has finished
- Streaming mode: Here the captured data is directly forwarded to the PC over USB, the internal RAM of U3Pro is only acting as transfer buffer, to avoid any buffer overflows due to slow USB as good as possible. In this mode the data can directly be displayed in a GUI

Out of legacy reasons (existing dslogic driver), the default behavior is buffered mode and unfortunately there is no way in pulseview to change it. The option to change it in the driver is there but pulseview is not showing this option, it is filtered in pulseview. If you use sigrok-cli you can switch between the two modes.

Actually, i'm currently working on some improvements in pulseview whereas I don't think they can be upstreamed as some of them are just quick hacks. The changes are:
- Performance improvement for storing samples in pulseview: Currently pulseview stores all samples in RAM in a uncompressed way. You will recognize this if you sample for instance at 100MHz for 10sec on 16 channels, it will occupy ~2GB of RAM (16Channels=2bytes * 10s*100M/s) - even if there is no single change in any of the signal. I changed this significantly in my version of pulseview, where I only store changes of the signal. So a 10sec capture at 100MHz with no changes in the signal consumes almost 0 RAM. The drawback was that I had to rewrite the complete logic on how to display samples at different zoom levels, so it's not perfect yet but i'm on it
- I have added support for switching between streaming mode and buffered mode for U3Pro. By default pulseview uses streaming mode. You can switch to buffered mode using the "Datalog" option in the device' config menu
- Support for quick interactive timing measure between two edges using the mouse. You click on one edge then you move the mouse and it shows the time in a hover popup
- Support for pulse/edge counting using mouse (both features are well known from DSView)

Latest development build is here: https://github.com/christianeisendle/pulseview/releases/tag/pulseview-0.5.4

If i find a bit more time, i might consider upstreaming few of those features once I cleaned up the implementation

dresco:
Thanks for your efforts on this. I couldn't open an issue on your repo, so commenting here..

I'm seeing the following issue on my DSLogic Plus, it just hangs at getting trigger. (The FPGA bitstream is the latest one from the DSView github).

```
LD_LIBRARY_PATH=~/sr/lib/ ~/sr/bin/sigrok-cli -d dreamsourcelab-dslogic -l 5 --samples 1000
sr: [00:00.000001] log: libsigrok loglevel set to 5.
sr: [00:00.000127] backend: libsigrok 0.6.0-git-23981f9/4:0:0.
sr: [00:00.000267] backend: Libs: glib 2.70.3 (rt: 2.70.3/7003:3), zlib 1.2.11, libzip 1.8.0, minilzo 2.10, libserialport 0.1.1/1:0:1 (rt: 0.1.1/1:0:1), libusb-1.0 1.0.24.11584 API 0x01000108.
sr: [00:00.000307] backend: Host: x86_64-pc-linux-gnu, little-endian.
sr: [00:00.000346] backend: SCPI backends: TCP, serial, USBTMC.
sr: [00:00.000375] backend: Firmware search paths:
sr: [00:00.000426] backend:  - /home/jone/.local/share/sigrok-firmware
sr: [00:00.000451] backend:  - /home/jone/sr/share/sigrok-firmware
sr: [00:00.000467] backend:  - /home/jone/.local/share/flatpak/exports/share/sigrok-firmware
sr: [00:00.000484] backend:  - /var/lib/flatpak/exports/share/sigrok-firmware
sr: [00:00.000499] backend:  - /usr/local/share/sigrok-firmware
sr: [00:00.000514] backend:  - /usr/share/sigrok-firmware
sr: [00:00.000598] backend: Sanity-checking all drivers.
sr: [00:00.000633] backend: Sanity-checking all input modules.
sr: [00:00.000658] backend: Sanity-checking all output modules.
sr: [00:00.000695] backend: Sanity-checking all transform modules.
srd: libsigrokdecode loglevel set to 5.
sr: [00:00.008491] dreamsourcelab-dslogic: Found a DSLogic device with API version 2
sr: [00:00.008525] hwdriver: Scan found 1 devices (dreamsourcelab-dslogic).
sr: [00:00.008563] device: dreamsourcelab-dslogic: Opening device instance.
sr: [00:00.008570] dreamsourcelab-dslogic: Firmware upload was not needed.
sr: [00:00.030222] dreamsourcelab-dslogic: Opened device on 1.43 (logical) / usb/1-3.1 (physical), interface 0, firmware 2.0, API version 2.
sr: [00:00.030248] dreamsourcelab-dslogic: Detected REVID=1, it's a Cypress CY7C68013A (FX2LP).
sr: [00:00.030272] dreamsourcelab-dslogic: Uploading FPGA firmware 'dreamsourcelab-dslogic-plus-fpga.fw'.
sr: [00:00.030354] resource: Attempt to open '/home/jone/.local/share/sigrok-firmware/dreamsourcelab-dslogic-plus-fpga.fw' failed: No such file or directory
sr: [00:00.030375] resource: Opened '/home/jone/sr/share/sigrok-firmware/dreamsourcelab-dslogic-plus-fpga.fw'.
sr: [00:00.041032] dreamsourcelab-dslogic: FPGA firmware upload done.
sr: [00:00.041537] hwdriver: sr_config_set(): key 50001 (limit_samples) sdi 0x54fc90 cg NULL -> uint64 1000
sr: [00:00.041642] session: Using thread-default main context.
sr: [00:00.041672] session: Starting.
sr: [00:00.041691] hwdriver: dreamsourcelab-dslogic: Starting acquisition.
sr: [00:00.042086] dreamsourcelab-dslogic: Configuring FPGA.
sr: [00:00.053545] dreamsourcelab-dslogic: No session trigger found
sr: [00:00.064703] dreamsourcelab-dslogic: Getting trigger.
```

christianeisendle:
Ok, my guess is that FPGA bitstream is still the old version, in fact, it seems that FW version (API version 1 and 2) and FPGA bitstream version are completely independent. So it seems that FPGA API is coupled to device while FW API is coupled to FW version. If this assumption is correct, the fix would be straight-forward since the FPGA related functionality is encapsulated in fpga_configure.
I will push the changes to a branch to test for you, before updating the PR

christianeisendle:
Actually, i just reviewed the code of DSView and i don't see any diversification in dsl_fpga_arm. (well, there is one but it seems to be unrelated)

I would propose to dump the content of "setting struct" in dsl_fpga_arm() while running DSView to a file and then load the content of the dump file in libsigrok dsl.c/fpga_configure() and send it there to the FPGA. Assuming this works, we just need to compare the content of setting struct between libsigrok and DSView.

Are you able to compile DSView under Linux? If so, i can give you a patch for DSView to dump the settings.

christianeisendle:
Hi @dresco,
thanks a lot for your support in debugging this issue. I just updated the PR and set now the GPIF to word-wide for non USB3.0 devices, as you suggested.
Let me know if it works out for you now.

Christian.

tobyp:
This also works for me, using the newest firmware from [1].

I suggest adding new entries to contrib/60-libsigrok.rules:

```
# DreamSourceLab DSLogic U3Pro16
ATTRS{idVendor}=="2a0e", ATTRS{idProduct}=="002a", ENV{ID_SIGROK}="1"
```

[1]: https://github.com/DreamSourceLab/DSView/blob/beb06ca1beab1c4f2d9da8b02fdd9b6639b93f0d/DSView/res/DSLogicU3Pro16.bin

christianeisendle:
> I suggest adding new entries to contrib/60-libsigrok.rules:

Thanks, wasn't aware about presence of the rules file - just added U3Pro16 and updated the PR

dresco:
Hi, just a quick observation. Have recently noticed that the pre-trigger capture ratio in pulseview isn't working as I would have expected..

For example, when set to 50% pre-trigger with 50M samples @ 10MHz, it only captures ~670ms of pre trigger data (instead of the expected 2.5s). The ratio seems dependent on the amount of data, 5M samples @ 1Mhz is okay, but at higher speeds (or longer duration) there is even less pre-trigger data stored. 

I think the distinction is that it works fine when the selected combination of data rate & duration is not using RLE. Just wondering if this is known / expected behaviour for these devices?

acassis:
> Hi, just a quick observation. Have recently noticed that the pre-trigger capture ratio in pulseview isn't working as I would have expected..
> 
> For example, when set to 50% pre-trigger with 50M samples @ 10MHz, it only captures ~670ms of pre trigger data (instead of the expected 2.5s). The ratio seems dependent on the amount of data, 5M samples @ 1Mhz is okay, but at higher speeds (or longer duration) there is even less pre-trigger data stored.
> 
> I think the distinction is that it works fine when the selected combination of data rate & duration is not using RLE. Just wondering if this is known / expected behaviour for these devices?

Hi @dresco and @christianeisendle what happened? Is this PR stalled because this issue with pre-trigger? I got a DSLogic U3Pro16 and want to use it with PulseView instead of DSView. Thank you very much!

christianeisendle:
Hi @dresco, @acassis,
apologize for the delayed response on this. I can confirm that there is indeed an issue with the pre-trigger. It seems when only 8 out of the 16 channels are activated it works but with all 16 channels i observe similar behavior as @dresco did. 
Tried with DSView and there it works fine, so it means it isn't an issue with the DSLogic firmware but there is a bug in the DSLogic sigrok driver.
Need some more days to debug this, will come back with an updated PR soon

acassis:
Hi @christianeisendle very nice! Sorry to hurry up you :-) At least it is working fine for up to 8 inputs.

dresco:
Hi @christianeisendle Thanks for picking this back up again. Just to confirm I can recreate this issue with fewer channels though (perhaps is more apparent with the lower hw spec of my DSLogic Plus)?

Here are some example images (triggering off channel 2, with 50% pre-trigger);

**10M samples @ 10MHz on 4 channels, 500ms pre-trigger data as expected**
image

**50M samples @ 50MHz on 4 channels, ~134ms pre-trigger data**
image

**50M samples @ 50MHz on 8 channels, ~67ms pre-trigger data**
image

And this pattern remains consistent, moving from 8 to 16 channels I get ~33ms of pre-trigger data. Changing the sampling rate also changes the pre-trigger duration (half the sample rate = double the pre-trigger time) etc. 

Has been a while since I first looked at this, but my suspicion at the time was that it happened if the selected sample rate and duration required RLE..

christianeisendle:
Hi @dresco,

thanks for sharing your findings. To me it really looks like a hardware limitation. I did a comparison with DSView to see how they handle it there. Interestingly, they don't allow changing pre-trigger in streaming mode at all, it's only supported in buffered mode.

Streaming mode (SR_CONF_CONTINUOUS in the driver) is enabled by default but back in January it wasn't possible to change that in pulseview. Coincidentally, @gsigh fixed that in Feb: https://github.com/sigrokproject/pulseview/commit/a00c1b6edc840564eea6bb22d292892970f0071e

So i built pulseview from latest master, disabled continous mode in device settings and after that pre-buffering works as expected:

grafik

Probably there are ways to fix it also for continous/streaming mode but since DSView also doesn't offer this option, i would be difficult to find out how to do it without reverse engineering the device. On the other hand, the guys from dreamsourcelab might have had a good reason to allow it only for buffered mode...

dresco:
Thanks for looking at this @christianeisendle.

I rebuilt with the latest upstream pulseview, and can confirm that it works somewhat better with continuous mode disabled. It does still reduce the amount of pre-trigger data as I increase the frequency or number of channels even further, but agree that it looks like a device limitation. 

This prompted me to take another look at DSView - and it shows the same behaviour of changing the amount of pre-trigger data (in buffer mode) as these settings are changed. Also, their UI logic seems a bit buggy, but DSView does actually attempt to adjust the maximum selectable trigger position as the frequency etc increases.

christianeisendle:
i can think of implementing this pre-buffering feature in the driver itself. For that we would need to reserve significant memory in the driver and store the pre-buffered data there (using a ring-buffer) and only if the driver detects an edge change it pushes the pre-buffer + the streaming data which comes after the trigger to the upper layer (pulseview/sigrok-cli). Currently pre-buffering is done in the device and since it has limited memory it will obviously start to limit the max. pre-buffer length depending on number of channels and sampling frequency.

I was just thinking about one option which could make sense: In case upper layer sets a prebuffer > 0, the driver implicitly enables buffered mode (with the drawback of having no live capturing after the has triggered). Let me know what you think about this option.

dresco:
> I was just thinking about one option which could make sense: In case upper layer sets a prebuffer > 0, the driver implicitly enables buffered mode (with the drawback of having no live capturing after the has triggered). Let me know what you think about this option.

I'd worry that  this might not be very intuitive?  Honestly, now knowing it's a device limitation (with similar results in vendor software), perhaps it just needs documenting somewhere as expected behaviour?

bkleiner:
i just tried the branch with my U3Pro16, it works beautifully! besides the documentation issue for the pre-trigger, is there something keeping this from being mainlined?

carlberg74:
Bought my self a U3Pro16 as xmas pressent a year or two ago and I have always wanted to use it with sigrok-cli, PulsView etc... and now I soon will!.. 
Wooohooo!  Excellent work guys! 
I wote for sending it to main!

dresco:
> I wote for sending it to main!

For whatever reason, many PR's appear stalled without upstream review, so you might do well to build from source (if you aren't already).

The [sigrok-util](https://github.com/sigrokproject/sigrok-util/) repo makes building fairly straightforward, just needs a small update to pull libsigrok from the right place. Fwiw, these are my changes to build for Linux..
```
diff --git a/cross-compile/linux/sigrok-cross-linux b/cross-compile/linux/sigrok-cross-linux
index 6ba7037..2b11f0c 100755
--- a/cross-compile/linux/sigrok-cross-linux
+++ b/cross-compile/linux/sigrok-cross-linux
@@ -69,6 +69,7 @@ cd $BUILDDIR
 GIT_CLONE="git clone --depth=1"
 
 REPO_BASE="git://sigrok.org"
+LIBSIGROK_REPO_BASE="-b dslogic_u3pro16 https://github.com/christianeisendle"
 
 # -----------------------------------------------------------------------------
 
@@ -85,7 +86,7 @@ cd ../..
 
 # libsigrok
 mkdir -p $PYPATH
-$GIT_CLONE $REPO_BASE/libsigrok
+$GIT_CLONE $LIBSIGROK_REPO_BASE/libsigrok
 cd libsigrok
 ./autogen.sh
 mkdir build
```

carlberg74:
Still not merged with main, I guess?

About building from source; I will need to build libsigrok to get the device to work in pulseview or do I also need to build pulseview?
(+ want to run pulseview on Windows)

Side note; someone should update the list of supported hardware.. this device still says planned, but that's not true... :D

christianeisendle:
Build for windows is supported via cross compile toolchain under Linux. 
This worked fine so far for me, I didn't manage to compile natively.

Just follow the instructions given in the README i  
http://sigrok.org/gitweb/?p=sigrok-util.git;a=tree;f=cross-compile/mingw

You may need to adjust the repository and/or branch in sigrok-cross-mingw

Am 23. November 2022 23:09:31 schrieb carlberg74 ***@***.***>:

ojg78:
Has there been any movement getting this reviewed? Seems like a shame to leave it here to fester.

dresco:
> Has there been any movement getting this reviewed? Seems like a shame to leave it here to fester.

fwiw I believe there is some effort going into triage of outstanding issues and PR's. There is an ongoing desire for more review input on code submissions.

As a random nobody, my thoughts on this PR are that it obviously works well for multiple people (myself included), but 
I think it would benefit from being rebased against current upstream, and the commit history being tidied up.

@christianeisendle don't know whether you still have the desire/time to pursue this? I don't mind having a go at it if not?

christianeisendle:
I feel we need to bring it to the official mailing list, not sure if the maintainers follow all pull-requests on github.
While i think DSLogig U3Pro16 works fine (aside from [this limitation with the prebuffer](https://github.com/sigrokproject/libsigrok/pull/170#issuecomment-1169173308) - which is a HW limitation) one open topic is testing with other DSLogic devices. I intentionally added the support into the existing dreamsourcelab-dslogic driver to avoid any code duplication but i'm unable to test with any other DSLogic devices. Would be good if somebody out with a legacy device could test this pull request.

Let me reach out to others via the mailing list

dresco:
> Let me reach out to others via the mailing list

Cool thanks. There seems to be more developer activity on the IRC channel, don't know whether that's another option for you?

nemanjan00:
Is anyone still working on this? 

Sigrok is also missing U3Pro32 support. I will have some time in the next few weeks to do some work on this.

nemanjan00:
It seems to be outputing data, but, only after 4min, 26sec, for some reason

mrv96:
@nemanjan00 any update?

nemanjan00:
Never managed to get it to work properly. Gave up. Just used slower LA I own with sigrok

MabezDev:
@christianeisendle Thanks for your work on this! As an owner of the V2 firmware DLogic Plus I'm excited to see that sigrok + V2 firmware is possible. Does this PR support other V2 devices as it stands, or would some modifications need to be made?

flisboac:
Is this PR still being worked on? I'm in the same situation as @MabezDev , I have a DSLogic Plus "Pango Variant" (which I believe is what everyone is referring to as "V2", please correct me if I'm wrong).

@christianeisendle Would this PR introduce support for that version? I see that there's the addition of a new firmware for the U3Pro16, but looking into the DSView repository, I see that there're at least two new versions of firmware for the DSLogic Plus "Pango" (I think):

- [DSView/res/DSLogicPlus-pgl12.bin](https://github.com/DreamSourceLab/DSView/commit/2532ad4c106f588c0e9e8dfbcb555644e992d4ad#diff-30016844c7a8cfe6eb376359af8cf48e223d969ba1b86890589a5476dcad4a12), later updated in commits [cfa0c1cb78f1d99fca9ae650ccae59cff40de4dd](https://github.com/DreamSourceLab/DSView/commit/cfa0c1cb78f1d99fca9ae650ccae59cff40de4dd#diff-30016844c7a8cfe6eb376359af8cf48e223d969ba1b86890589a5476dcad4a12) and [63955d72939dc07e38a11c7d87918fcd32411054](https://github.com/DreamSourceLab/DSView/commit/63955d72939dc07e38a11c7d87918fcd32411054#diff-30016844c7a8cfe6eb376359af8cf48e223d969ba1b86890589a5476dcad4a12)
- [DSView/res/DSLogicPlus-pgl12-2.bin](https://github.com/DreamSourceLab/DSView/commit/673468397132cf642cfaf337b204faf85acf89dc#diff-3400710c96003aad1b0360ed4b9909f9461323308420245561f54b97d0084b4b), which I believe is for the API V2, perhaps, given the name. (I don't still understand what this entails, exactly, but AFAICT, it's a software/firmware version, distinct from the hardware. Is that it?)

Would it be sufficient to arrange the code so that the correct firmware is referenced and uploaded, or is there something else to consider?

cantclosevi:
Hello People,

the API check checks if the USB Device has "Instrument v2" in its ID... the Xilinx FPGA variants are also API V2 devices.

The guys from DreamSourceLabs added a [security check](https://github.com/DreamSourceLab/DSView/blob/master/libsigrok4DSL/hardware/DSL/dsl.c#L2681) for newer Devices
The security check seems to be necessary for the device activation.

I forked @christianeisendle libsigrok repo and added the security check to support to Pango FPGA Devices (0x030, 0x0031, 0x0034 and 0x0035).

My fork can be found [here](https://github.com/cantclosevi/libsigrok)
The .fw files with my naming scheme can be found [here](https://github.com/cantclosevi/sigrok-firmware)

Cheers

AreYouLoco:
> Hello People,
> 
> the API check checks if the USB Device has "Instrument v2" in its ID... the Xilinx FPGA variants (0x0020 and 0x0021) are also API V2 devices.
> 
> The guys from DreamSourceLabs added a [security check](https://github.com/DreamSourceLab/DSView/blob/master/libsigrok4DSL/hardware/DSL/dsl.c?rgh-link-date=2024-06-21T09%3A18%3A40Z#L2681) for newer Devices The security check seems to be necessary for the device activation.
> 
> I forked @christianeisendle libsigrok repo and added the security check to support the Pango FPGA Devices (0x030, 0x0031, 0x0034 and 0x0035).
> 
> My fork can be found [here](https://github.com/cantclosevi/libsigrok) The .fw files with my naming scheme can be found [here](https://github.com/cantclosevi/sigrok-firmware)
> 
> Cheers

--- BEGIN OFF-TOPIC ---
I wanted to say thank you to @cantclosevi for the support for DSLogic Plus v2 (with Pango chipset) I just pulled your fork rebased agains upstream master and build. And then replaced symlinks in the system to point to freshly build. Plus firmware with your naming convention and it works!. I mean at least PulseView detects the device and the green light lids so the firmware is loaded. Thanks again. I hope it works as an analyzer now. This should be merged upstream for sure ASAP.

PulseView_DSLogic_Plus_v2
Here is screenshot-proof. Where may I reach you to donate some $ to you and to @christianeisendle as well. Thank you guys for your work. That's it.

--- END OFF-TOPIC ---

Sfinx:
Any merging news ? Interested in U3Pro32 (2a0e:002c) support

sharkcz:
IANAL, there are better licenses for distributable firmware files than MIT, but I won't object to the selection of MIT here :-)

PR #171, opened 2022-01-13
input/vcd: add parsing support for SV 'logic' type

7FM:
Hi,
my use case is "simple":
I have SystemVerilog (SV) code with Verilator simulation that produces USB signals.
Verilator can create VCD traces that I want to import via PulseView and check with the awesome decoder implementations if everything is sound.
However, as I am using SV, I have primarily variables of type `logic` which the VCD parser does not recognize.
Hence, I need to manually edit my VCD traces to change the `$val` types from `logic` to `wire` or `reg` which is somewhat inconvenient.
Took me some time to figure out what's the problem here, as the only response that I got from PulseView was `No channels enabled.`.
The changes to add parsing support for the `logic` type is straight forward and should'nt cause problems.

Also, here is some minimal example of an VCD file for i.e. regression tests:
```
$scope module doesntMatter $end
$var logic 1 " D0 $end
$upscope $end
$enddefinitions $end
#1
$dumpvars
1"
$end
```

### DIFF
--- a/src/input/vcd.c
+++ b/src/input/vcd.c
@@ -45,7 +45,7 @@
  * Based on Verilog standard IEEE Std 1364-2001 Version C
  *
  * Supported features:
- * - $var with 'wire' and 'reg' types of scalar variables
+ * - $var with 'wire', 'reg' and 'logic' types of scalar variables
  * - $timescale definition for samplerate
  * - multiple character variable identifiers
  * - same identifer used for multiple signals (identical values)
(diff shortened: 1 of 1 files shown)


biot:
Sorry this didn't get a proper review earlier.

There is no problem with your code, but I'm not having any luck finding the `logic` variable type. I checked IEEE 1364-2005 and there are a bunch of variable types but not logic. Is this a nonstandard Verilator thing? Does anything else support this?

7FM:
`logic` is a **System**Verilog data type which is an extension of Verilog. For SystemVerilog you would have to check IEEE 1800-2017.

biot (approved):

abraxa:
Merged as https://sigrok.org/gitaction?p=libsigrok.git;a=commit;h=47a98d305c6ad9b38c302618d864f1e77e9f036a, thank you for contributing!

For future reference, please note that we don't do merges, only rebases. Merge commits require manual fixup on my part, so it's best to avoid them :)

PR #172, opened 2022-01-16
RFC: Preliminary support for Française d'Instrumentation AMS515

mmuman:
[I got this old lady last week](https://twitter.com/mmu_man/status/1482158039604080651) when cleaning up at the university, and managed to figure out the protocol.

I'm not sure where to go now, so seeking input. I tried to come up with sensible code by looking at other drivers but I probably overlooked some stuff.

- Driver & file naming is challenging, due to accented char and apostrophe (`d'` means `of`) in the name, I think it's the best compromise.
- The front panel can be disabled, but it also shuts down the display, making it appear powered off, so not sure it's desirable.
- I can't make much of it in SmuView, the panels just stay with dashes everywhere, maybe it expects more possible configurations, or maybe I need to implement monitoring to report configuration changes ?
- Oddly it seems to remember settings when powered on, but only when set from the front panel, not by the serial port.
- It's possible to change the selected output on the front panel display, don't know what to do with that.
- Outputs are always on, should I make it appear in the config or not.
- I haven't tried yet to see how overcurrent is reported.


mmuman:
Setting target voltage now works properly, some rounding errors with sigrok-cli, but it works fine in SmuView now. Changing a channel makes it selected on the front panel so the display is also updated.

mmuman:
I used a hack to get a timer without depending on timerfd which seems to be a completely non-POSIX Linuxism I'd rather avoid (since I intend to possibly use it in Haiku as well), with an invalid fd as GIOChannel, but I didn't test to make sure it works in Windows. I might have a try with mingw someday but I only have an XP VM around.

mmuman:
Still open for comments but I consider it ready to merge on my side.

mmuman:
Another option for the timer would be to use a `pipe()` but that'd probably fail on Windows.

mmuman:
Rebased and added call to sr_channel_group_new.

src/hardware/francaise-instrumentation-ams515/api.c
```diff
@@ -52,14 +52,16 @@ static const uint32_t drvopts[] = {
 
 static const uint32_t devopts[] = {
 	SR_CONF_CHANNEL_CONFIG | SR_CONF_GET | SR_CONF_SET | SR_CONF_LIST,
+	//SR_CONF_OVER_CURRENT_PROTECTION_ENABLED | SR_CONF_GET,
+	//SR_CONF_OVER_CURRENT_PROTECTION_ACTIVE | SR_CONF_GET,
```
acassis, line 56: Why did you comment out the OC options?

mmuman: IIRC I was trying to get SmuView to report the OC condition properly so I tried several ways but it didn't work, always reported on the first channel regardless which was shorted. 
I should try again now that it's not used as a base for my external screen anymore.

mmuman:
Rebased and did some cleanup.
I enabled back OCP reporting at device level, and fixed it when used with `sigrok-cli` as it seems the monitoring doesn't get setup, at least the condition wasn't reported like in SmuView.

The pending issues are:
- OCP reporting doesn't tell which channel (Are we missing a `cg_session_send_meta` maybe?)
- The weird use of `sr_session_source_add_channel` with a -1 Unix fd for timing, which I don't know if it works on non-unix platforms. I should try with mingw someday.

biot:
Sorry this did not get reviewed earlier. I'd like to help get this reviewed/approved, if you still want this driver merged.

A few preliminary notes:
- Please squash this into two commits: one for the driver skeleton, and the next for your code filling it out.
- Why are you using this GIOChannel? You can put a timeout on the serial port just the same, and that's definitely portable.
- It's very odd that this driver doesn't continually get or fetch supplied voltage/current, I don't think there's any other PPS driver that doesn't. Is there really no way to get this info from the hardware? Maybe one of the `'S'` commands -- surely the display shows what it's putting out? It's not a dealbreaker -- there is still value in controlling the instrument via sigrok -- but querying current output should definitely be there if possible.
- Please remove debug/TODO/xxx comments, the driver should be in fully working state when submitted to mainline. It doesn't need to have all the features yet, but what's there should be complete.
- You're not the first to wonder how to report per-channel conditions like OCP -- it came up in PR #101 as well. We'll tackle this.
- Lots of coding style fixes needed, please check https://github.com/sigrokproject/libsigrok/blob/master/HACKING

mmuman:
Sadly no, it only reports over-current condition and that's it, and it stops responding until the condition is suppressed.
I'll try to have a look again later on when the electronics lab is installed here.

mmuman:
I just rebased for now, I'll have to address the review next. Also I had to rebase over #253 as it wouldn't build on Debian Sid anymore, so we probably want to merge that one first.

> A few preliminary notes:
> 
> * Please squash this into two commits: one for the driver skeleton, and the next for your code filling it out.

I'll do this last, history helps fixing stuff.

> 
> * Why are you using this GIOChannel? You can put a timeout on the serial port just the same, and that's definitely portable.

I don't remember exactly, I think that's the only example code I found in the acme driver that mentioned a timer, but yeah I could just follow the pce-322a driver here and call `serial_source_add`.

> 
> * It's very odd that this driver doesn't continually get or fetch supplied voltage/current, I don't think there's any other PPS driver that doesn't. Is there really no way to get this info from the hardware? Maybe one of the `'S'` commands -- surely the display shows what it's putting out? It's not a dealbreaker -- there is still value in controlling the instrument via sigrok -- but querying current output should definitely be there if possible.

Well it's quite primitive. I can get the updated target if someone changes it at the panel but that's it. At least from the commands I could find. I'll try to contact the manufacturer as it seems they're still around but I wonder if they even still have documentation about it.

> * Please remove debug/TODO/xxx comments, the driver should be in fully working state when submitted to mainline. It doesn't need to have all the features yet, but what's there should be complete.

I'll look through these.

> * You're not the first to wonder how to report per-channel conditions like OCP -- it came up in PR [new SR_CONF_CHANNEL_GROUP meta package #101](https://github.com/sigrokproject/libsigrok/pull/101) as well. We'll tackle this.

Ah great, I'll try to handle the rest meanwhile. I guess we can leave that TODO.

Now, I just noticed it actually accepts target updates for the other channels when overcurrent, though the regulation seems to be broken, fluctuating by like 1V… I think it's because the MCU is busy blinking the overcurrent indicator on the LED display. I wouldn't trust it for serious stuff.

> 
> * Lots of coding style fixes needed, please check https://github.com/sigrokproject/libsigrok/blob/master/HACKING

Ah, I usually try to follow whatever convention I see around but I wrote most of that at night, so…

mmuman:
Let me know if there are other concerns, then I'll squash that.
There are several of them at the uni, I can test them to make sure their firmware is all 4.1.

mmuman:
Just checked with the 9 others at the Uni, they are all at 4.1.

PR #173, opened 2022-01-29
input/vcd: add option to manually specify & enforce a samplerate

7FM:
This PR is somewhat related to my other PR https://github.com/sigrokproject/libsigrok/pull/171, but independent!

My motivation for this feature is that cycle based simulation tools don't really care about the actual clock frequency and produce VCD dumpy with each time step corresponding to i.e. a clock change. If you want to use the created dump with pulseview then the `timescale` section is used to determine the period and to calculate the sample rate. In this case this value has not only nothing to do with the desired clock frequency (or sample rate) but it might not even be possible to represent the desired frequency with this method. This is due to the usage of integers to represent the period and time steps which can not as simple frequencies as 30 MHz (1/30 MHz = 33.333... µs as period). Hence, even for some simple clock signals the existing `timescale` field is not enough. Therefore, a new option was added to be able to overwrite the desired sample rate when importing a VCD trace. Another solution would have been the introduction of a custom section within the VCD file. I am not familiar with the VCD spec but I would expect this to not be standard compliant and needing more code modifications than the simple addition of another option. 

But if desired and preferred over the additional import option, then I am willing adjust the code to parse a custom section that can specify the sample rate directly instead of the period.

Oh and btw, I have tested this PR in combination with my other PR https://github.com/sigrokproject/libsigrok/pull/171 to analyze simulated USB interactions while using the protocol decoders.

### DIFF
--- a/src/input/vcd.c
+++ b/src/input/vcd.c
@@ -35,6 +35,9 @@
  *     timestamp 0.
  *   Value > 0: Start at the given timestamp.
  *
+ * samplerate: overwrite/manually specify the samplerate for this VCD file
+ *   and ignore timescale sections.
+ * 
  * downsample: Divide the samplerate by the given factor. This can
  *   speed up operation on long captures.
  *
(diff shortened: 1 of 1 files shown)


biot (approved):

abraxa:
Merged as https://sigrok.org/gitaction?p=libsigrok.git;a=commit;h=02c51d8a5f71152a5b230fc463719c13c9404f44, thank you for contributing!

PR #174, opened 2022-02-08
enables the "external clock" of  Zeroplus HW driver and fixes a hidden bug

g-user:
This enables the "external clock mode" in the Zeroplus LAP-C Logic Cube driver (falling and rising edge), 
so that this mode also can also be used in sigrok-cli and pulseview
in the same code a hidden bug is corrected: the internal clock is chosen by

/* Sele_Inside_Outside_Clock */
gl_reg_write(devh, CLOCK_SOURCE, 0x01);

and not (original code in analyzer.c):

/* Sele_Inside_Outside_Clock */
gl_reg_write(devh, CLOCK_SOURCE, 0x03);

this was proven by "USB-sniffing" the vendor software

### DIFF
--- a/src/hardware/zeroplus-logic-cube/analyzer.c
+++ b/src/hardware/zeroplus-logic-cube/analyzer.c
@@ -462,7 +462,7 @@ SR_PRIV void analyzer_configure(libusb_device_handle *devh)
 	gl_reg_write(devh, MEMORY_LENGTH, g_memory_size);
 
 	/* Sele_Inside_Outside_Clock */
-	if (g_clock_ext==0)
+	if (g_clock_ext == 0)
 		gl_reg_write(devh, CLOCK_SOURCE, 0x01);
 	else if (g_edge==0)
 			gl_reg_write(devh, CLOCK_SOURCE, 0x02);
--- a/src/hardware/zeroplus-logic-cube/analyzer.c
+++ b/src/hardware/zeroplus-logic-cube/analyzer.c
@@ -114,6 +114,9 @@ static int g_trigger_count = 1;
 static int g_filter_status[8] = { 0 };
 static int g_filter_enable = 0;
 
+static int g_clock_ext=0;
+static int g_edge=0;
+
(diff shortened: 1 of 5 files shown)


src/hardware/zeroplus-logic-cube/analyzer.c
```diff
@@ -459,7 +462,12 @@ SR_PRIV void analyzer_configure(libusb_device_handle *devh)
 	gl_reg_write(devh, MEMORY_LENGTH, g_memory_size);
 
 	/* Sele_Inside_Outside_Clock */
-	gl_reg_write(devh, CLOCK_SOURCE, 0x03);
+	if (g_clock_ext==0)
```
acassis, line 465:
```suggestion
	if (g_clock_ext == 0)

File: src/hardware/zeroplus-logic-cube/analyzer.c [hunk 2]
```diff
@@ -459,7 +462,12 @@ SR_PRIV void analyzer_configure(libusb_device_handle *devh)
...
+		gl_reg_write(devh, CLOCK_SOURCE, 0x01);
+	else if (g_edge==0)
```
acassis, line 467:
```suggestion
	else if (g_edge == 0)

acassis:
@g-user it is important to keep spacing as in the original code. I think Sigrok needs a CI with style code check like we do for NuttX RTOS.

g-user:
changed the spacing as suggested

fenugrec:
Some notes as an ordinary observer:

- the if + elseif + else construction should probably have braces everywhere.
- the whitespace fixups should be squashed into the first commit. Also I think you missed one or two instances of missing-space-around-assignment.
- the commit message is probably not acceptable as-is: see log for examples, first line looks like `<driver-name>: <short description>` , then empty line, then more details.

abraxa:
Merged as https://sigrok.org/gitaction?p=libsigrok.git;a=commit;h=0dfa850efb5553940982633ab96a8612860e87b4, thank you!

PR #114, opened 2021-01-23
siglent-sds: original driver fixes, USBTMC support

voneiden:
NOTE: Needs testing by someone who owns an older siglent scope.

Fixes a few memory leaks, one inconsistency with DATA_SOURCE and adds USBTMC support.

Addresses bugs

* https://sigrok.org/bugzilla/show_bug.cgi?id=1628 (usbtmc/eseries)
* https://sigrok.org/bugzilla/show_bug.cgi?id=1130 (usbtmc)

I've split off the eseries changes into a separate pull request #118  as that code should not require testing by someone with an older scope model.

PR #176, opened 2022-02-17
Add support for Siglent's SDS2000X Plus series

gralco:


src/hardware/siglent-sds/protocol.c
```diff
@@ -322,6 +322,7 @@ static int siglent_sds_read_header(struct sr_dev_inst *sdi)
 	ret = sr_scpi_read_data(scpi, buf, SIGLENT_HEADER_SIZE);
 	if (ret < SIGLENT_HEADER_SIZE) {
 		sr_err("Read error while reading data header.");
+		sr_dbg("Device returned %i bytes.", ret);
```
gralco, line 325: This line was added when I was trying to debug why capturing a waveform wasn't working over USB. Doing this showed that it was only getting 52 bytes when done over USB. This line should hopefully help with future debugging.

gralco:
This was tested against tag 0.5.2 with 3decd3b1f0cbb3a035f72e9eade42279d0507b89 patch on top of it.

src/hardware/siglent-sds/protocol.c
```diff
@@ -567,7 +568,7 @@ SR_PRIV int siglent_sds_receive(int fd, int revents, void *cb_data)
 				} else {
 					sr_dbg("Requesting: %" PRIu64 " bytes.", devc->num_samples - devc->num_block_bytes);
 					len = sr_scpi_read_data(scpi, (char *)devc->buffer, devc->num_samples-devc->num_block_bytes);
```
gralco: This line here is the culprit. If I remove the subtraction devc->num_block_bytes then I get the entire waveform, no points missing.

gralco:
I just added support for the digital (logic analyzer) channels as well.

gralco:
@uwehermann If you get a chance, could you please take a look at this merge request. I believe it's ready to be merged.

gralco:
@gsigh Could you help get this merged?

src/hardware/siglent-sds/protocol.c
voneiden: FYI see these findings on using USB on SDS1104X-E: https://github.com/sigrokproject/libsigrok/pull/114/files#diff-b2e40d0d92f027de6ccfdb0e8581aca4192f2408ec3eb900fcc1a6520d054fd3R39

The USB packet size is 64 bytes with a 12 byte packet header so you get 52 bytes of data per packet. Additionally at least the SDS1104X-E has an internal buffer of 61440 bytes (variable named show_send_buffer_size can be seen on uboot log) which gets refilled only when it has been completely read. And while it's refilling the read will fail.

gralco:
My last push just added some fixes to work with the latest firmware (V1.5.2R1).

gralco:
@abraxa @fenugrec @gsigh Could you please have a look at this PR? It would be nice to get it merged upstream.

src/hardware/siglent-sds/api.c
```diff
@@ -429,6 +437,36 @@ static int config_get(uint32_t key, GVariant **data,
 			*data = g_variant_new_string("History");
 		break;
 	case SR_CONF_SAMPLERATE:
+		/* SDS2000X+: The channels need to be enabled/disabled prior to
+		 * obtaining the sample rate when the data source is set to display.
+		 * If a channel is turned on/off on the oscilloscope then use Sigrok's settings instead. */
+		if (!strcmp(devc->model->series->name, "SDS2000X+") && devc->data_source == DATA_SOURCE_SCREEN) {
```
fenugrec, line 443: I prefer avoiding to test things like model name strings for feature support... adding a bool flag (or bit-masked enums, I would look at other more mature drivers to see if there is a more common pattern) would be preferable esp. if future models also need this

src/hardware/siglent-sds/api.c
```diff
@@ -429,6 +437,36 @@ static int config_get(uint32_t key, GVariant **data,
...
+					char *cmd = g_strdup_printf("C%d:TRA?", ch->index + 1);
+					sr_scpi_get_string(sdi->conn, cmd, &channel_state);
+					if (!strcmp(channel_state, "ON") && !ch->enabled)
+						if (siglent_sds_config_set(sdi, "C%d:TRA %s", ch->index + 1,
+							ch->enabled ? "ON" : "OFF") != SR_OK)
+							return SR_ERR;
+					if (!strcmp(channel_state, "OFF") && ch->enabled)
+						if (siglent_sds_config_set(sdi, "C%d:TRA %s", ch->index + 1,
+							ch->enabled ? "ON" : "OFF") != SR_OK)
+							return SR_ERR;
```
fenugrec, line 458: this is getting into some pretty deep indent levels, maybe time to split some functionality into a separate func ?

src/hardware/siglent-sds/api.c
```diff
@@ -429,6 +437,36 @@ static int config_get(uint32_t key, GVariant **data,
...
+		 * If a channel is turned on/off on the oscilloscope then use Sigrok's settings instead. */
+		if (!strcmp(devc->model->series->name, "SDS2000X+") && devc->data_source == DATA_SOURCE_SCREEN) {
+			GSList *l;
+			for (l = sdi->channels; l; l = l->next) {
+				ch = l->data;
+				if (ch->type == SR_CHANNEL_ANALOG) {
+					char *channel_state;
+					char *cmd = g_strdup_printf("C%d:TRA?", ch->index + 1);
+					sr_scpi_get_string(sdi->conn, cmd, &channel_state);
+					if (!strcmp(channel_state, "ON") && !ch->enabled)
```
fenugrec, line 451: not checking if sr_scpi_get_string() failed before strcmp (i didnt check if it sets channel_state to null or to an empty string if it fails)

src/hardware/siglent-sds/api.c
```diff
@@ -857,10 +895,12 @@ static int dev_acquisition_start(const struct sr_dev_inst *sdi)
...
 	// devc->digital_frame_size = devc->model->series->buffer_samples;
 
-	siglent_sds_get_dev_cfg_horizontal(sdi);
+	if (strcmp(devc->model->series->name, "SDS2000X+"))
+		siglent_sds_get_dev_cfg_horizontal(sdi);
+	siglent_sds_get_dev_cfg_vertical(sdi);
 	switch (devc->model->series->protocol) {
 	case SPO_MODEL:
-		if (siglent_sds_config_set(sdi, "WFSU SP,0,TYPE,1") != SR_OK)
+		if (siglent_sds_config_set(sdi, "WFSU SP,0,NP,0,FP,0") != SR_OK)
```
fenugrec, line 903: you are changing this for all models , is this intended ?

src/hardware/siglent-sds/protocol.c
```diff
@@ -167,11 +167,18 @@ SR_PRIV int siglent_sds_capture_start(const struct sr_dev_inst *sdi)
...
+				float wait = devc->memory_depth_analog / 100;
+				sr_spew("Waiting %.f ms for the instrument to enter ARM mode.", wait / 1000);
+				g_usleep(wait);
+			}
 			if (sr_scpi_get_string(sdi->conn, ":INR?", &buf) != SR_OK)
 				return SR_ERR;
 			sr_atoi(buf, &out);
 			g_free(buf);
-			if (out == DEVICE_STATE_TRIG_RDY) {
+			if (out == DEVICE_STATE_TRIG_RDY || out == DEVICE_STATE_STOPPED) {
```
fenugrec, line 181: another change that impacts other devices, would like to hear rationale, testing on other models to confirm no breakage, and separate commit unless it's directly related to sds2000 support

src/hardware/siglent-sds/protocol.c
```diff
@@ -283,9 +290,23 @@ SR_PRIV int siglent_sds_channel_start(const struct sr_dev_inst *sdi)
 	switch (devc->model->series->protocol) {
 	case NON_SPO_MODEL:
 	case SPO_MODEL:
-		s = (ch->type == SR_CHANNEL_LOGIC) ? "D%d:WF?" : "C%d:WF? ALL";
-		if (sr_scpi_send(sdi->conn, s, ch->index + 1) != SR_OK)
-			return SR_ERR;
+		if (!strcmp(devc->model->series->name, "SDS2000X+")) {
```
fenugrec, line 293: another strcmp with model name

src/hardware/siglent-sds/protocol.c
```diff
@@ -315,13 +336,25 @@ static int siglent_sds_read_header(struct sr_dev_inst *sdi)
 	struct dev_context *devc = sdi->priv;
 	char *buf = (char *)devc->buffer;
 	int ret, desc_length;
-	int block_offset = 15; /* Offset for descriptor block. */
+	int block_offset; /* Offset for descriptor block. */
+	if (!strcmp(devc->model->series->name, "SDS2000X+"))
```
fenugrec, line 340: maybe would make sense to have this value (and header_size just below) in the device descriptor struct. Less code.

src/hardware/siglent-sds/protocol.c
```diff
@@ -364,76 +411,127 @@ static int siglent_sds_get_digital(const struct sr_dev_inst *sdi, struct sr_chan
...
-			for (int ii = 0; ii < 8; ii++, sample >>= 1) {
-				if (ch->index < 8) {
-					channel_index = ch->index;
-					if (data_low_channels->len <= samples_index) {
-						tmp_value = 0; /* New sample. */
-						low_channels = TRUE; /* We have at least one enabled low channel. */
-					} else {
-						/* Get previous stored sample from low channel buffer. */
-						tmp_value = g_array_index(data_low_channels, uint8_t, samples_index);
+		if (ch->type == SR_CHANNEL_LOGIC) {
```
fenugrec, line 414: this section is complex and hard to review (I am skipping over it at this time)
Also if there's that many different sub-cases with "if model is SDS2000", it starts making sense to have a separate function entirely.

src/hardware/siglent-sds/protocol.c
```diff
@@ -519,10 +617,12 @@ SR_PRIV int siglent_sds_receive(int fd, int revents, void *cb_data)
...
 			case SPO_MODEL:
-				/* The older models need more time to prepare the the output buffers due to CPU speed. */
-				wait = (devc->memory_depth_analog * 2.5);
-				sr_dbg("Waiting %.f0 ms for device to prepare the output buffers", wait / 1000);
-				g_usleep(wait);
+				if (strcmp(devc->model->series->name, "SDS2000X+")) {
+					/* The older models need more time to prepare the the output buffers due to CPU speed. */
+					wait = (devc->memory_depth_analog * 2.5);
+					sr_dbg("Waiting %.f ms for device to prepare the output buffers", wait / 1000);
+					g_usleep(wait);
```
fenugrec, line 624: I always dislike  sleep/wait calls. Is there no way to tweak SCPI read timeouts ?

src/hardware/siglent-sds/protocol.c
```diff
@@ -609,16 +727,18 @@ SR_PRIV int siglent_sds_receive(int fd, int revents, void *cb_data)
 					g_array_free(data, TRUE);
 				}
 				len = 0;
-				if (devc->num_samples == (devc->num_block_bytes - SIGLENT_HEADER_SIZE)) {
+				if (devc->num_samples == devc->num_block_bytes) {
```
fenugrec, line 730: changing common code for non-obvious reason

fenugrec (commented):
See comments.
-  I have issues with some of the changes that would seem to impact the other devices ? If that is the case, they belong in a separate commit with a good explanation and testing that confirms no breakage
- lots of strcmp with the model name; in some areas it looks like there is so much difference that it would  make sense to split code into separate functions.
- I didn't even look at code that looked like for{if{do{if{if{for{if{if... that is way too many indents and almost unreviewable
- you replaced some hardcoded constants with something that ends up in the device struct - that is great, but could/should have been a separate commit for those cases that could be applied independantly of adding SDS2000 per se. Small changesets are much prefered.

PR #177, opened 2022-02-21
Add Korad KD3005P version 4.1 firmware support

Rendman:
This PR adds support for the 4.1 version firmware of the Korad KD3005P power supply. This change was tested with sigrok-cli and SmuView and all features are present and functional.


knarfS:
This PR can be closed, power supply was added with commit d154c6b

PR #178, opened 2022-02-28
initial raspbery pi pico

pico-coder:


pico-coder:
Thanks for the feedback, I misunderstood the comments about the dual pull request.

I just did a fresh install of xemacs on my rpi and looks like I need to get its tab/spaces setup right.

As for the braces/parents that's just my weakness that I code C about once every 10 years and always forget the operator precedence related to pointers.

PR #91, opened 2020-09-18
Rigol DG series signal generator driver.

tjko:
Initial Driver implementation with support for the DG1000z series signal generators:

DG1022Z
DG1032Z
DG1062Z

Wiki entry created: https://sigrok.org/wiki/Rigol_DG1000z_Series

Also started page for Signal Generators: https://sigrok.org/wiki/Programmable_signal_generator

(replaces #87)

PR #179, opened 2022-03-01
raspberrypi-pico: Initial driver skeleton.

pico-coder:
hopefully I got it right this time.  I'll submit the 2nd pull request once this is good.


jr-oss:
Just the skeleton probably won't go into mainline (what would it be good for?).
You should continue adding your changes and start a PR once it is complete.
Like this PR, e.g.: https://github.com/sigrokproject/libsigrok/pull/91

If you need a special setup or firmware running on the pico, you should describe this in the wiki on sigrok.org, as well.
For providing firmware see also https://sigrok.org/wiki/Firmware

pico-coder:
ok.  When I read this I was thinking a patch and a pull request were the same thing.
https://github.com/sigrokproject/libsigrok/blob/master/HACKING

_That means your patch submission later will consist of at least two patches: the initial one
adding the skeleton driver, and one or more additional patches that actually
implement the respective driver code._

PR #182, opened 2022-03-10
PeakTech 6070/6075 support

ensc:
Patchset provides a new driver for the PeakTech 607x power supplies.  

**NOTE**: only 6070 has been tested; the 6075 implementation is based on documentation and reverse engineering.


acassis (approved):

src/hardware/peaktech-607x/api.c
```diff
@@ -0,0 +1,1410 @@
...
+		default:
+			return SR_ERR_NA;
+		}
+	} else {
+		/* handle channel group configuration options */
+		struct sr_channel const			*ch = cg->channels->data;
+		size_t					idx = ch->index;
+		struct peaktech_chan_config const	*cfg;
+
+		g_assert(idx < devc->model->num_chan);
```
Lightwars, line 542: Right before g_assert(idx):
`if(devc->model->model == PEAKTECH_MODEL_6075 && idx == 2) { idx = 1; }`

src/hardware/peaktech-607x/api.c
```diff
@@ -0,0 +1,1410 @@
...
+			return SR_ERR_NA;
+		}
+	} else {
+		/* handle channel group configuration options */
+		struct sr_channel const			*ch = cg->channels->data;
+		size_t					idx = ch->index;
+		struct peaktech_chan_config		*cfg;
+		double					dval;
+
+		if (idx >= ARRAY_SIZE(devc->config))
```
Lightwars, line 632: Right before if idx:
`if(devc->model->model == PEAKTECH_MODEL_6075 && idx == 2) { idx = 1; }`

Lightwars (changes requested):
I have tested your changes with a P6075 it works for channel group CH1, but if I try to set the voltage or current for CH2 it fails.

So I have found out that idx was set to 2 and I added the mentioned codeline to make it work.
But at the moment I can not see why the channel group data index is 0 for CH1 and 2 for CH2, this must be the root cause.

src/hardware/peaktech-607x/api.c
ensc: thanks for testing it.  Fix should be

```c
		size_t					idx = ch->index / 2;
```

Code tries to find out the channel configuration and reads the index of the first channel of a channel group.  As there are two such channels per group (see _scan() function below) it must be divided by 2.

ensc: I guess, I can refer the cfg directly in `cg->priv` without using the index....

Lightwars: > thanks for testing it. Fix should be
> 
> ```c
> 		size_t					idx = ch->index / 2;
> ```
> 
> Code tries to find out the channel configuration and reads the index of the first channel of a channel group. As there are two such channels per group (see _scan() function below) it must be divided by 2.

Okay, that makes sense. I didn't know it works like this, but had some thoughts in that direction.
Let me know if I should test something more with the P6075.

ensc: I updated the patchset; thanks again for testing.

Lightwars (changes requested):

src/hardware/peaktech-607x/api.c
```diff
@@ -0,0 +1,1397 @@
...
+ *  will be returned.
+ */
+static enum recv_result peaktech_recv(struct peaktech_device *devc,
+				      int revent, size_t max_len)
+{
+	struct sr_dev_inst const	*sdi = devc->sdi;
+	int				len;
+	size_t				send_len;
+
+	g_assert(max_len < sizeof devc->recv_buf);
```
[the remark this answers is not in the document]
ensc: good catch with the `g_assert()`.

Can you try

```diff
--- a/src/hardware/peaktech-607x/api.c
+++ b/src/hardware/peaktech-607x/api.c
@@ -1207,7 +1207,7 @@ static void pt_report_init_common(struct peaktech_device *devc,
 {
        report->analog = (struct sr_datafeed_analog) {
                .data           = report->data,
-               .num_samples    = devc->model->num_chan,
+               .num_samples    = 1,
                .encoding       = &report->encoding,
                .meaning        = &report->meaning,
                .spec           = &report->spec,
```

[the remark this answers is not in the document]
Lightwars: If have tested you suggestion and I looks good now. Thanks.

I haven't noticed that my writings weren't posted, so sorry about getting back late with the g_assert and the other stuff.

biot (changes requested):
Sorry this driver didn't get a proper review earlier. I hope you still want to get this merged. However I'm afraid it needs quite a lot of work.

- We like drivers to follow a particular pattern -- `api.c` for driver registration and its callbacks, `protocol.c` for hardware-specific code, and `protocol.h` for definitions needed by both. This helps us maintain drivers long after the original author has gone. Your driver is rather more freeform. Please use the `new-driver` script in the sigrok-util repo to generate your driver skeleton patch, then a second patch with your actual code. You'll see these two patches in the git repo for just about any new driver.
- We have a specific coding style in the project, and need contributors to follow it. Even stuff that isn't explicitly spelled out -- lining up variable and structs with tons of whitespace, say -- should be obvious from a glance at any other driver. Please read this doc: https://github.com/sigrokproject/libsigrok/blob/master/HACKING
- There is no need to handle all kinds of endianness in your driver. Even when you do -- if the hardware sends out data with an endianness implied -- there are lots of internal macros to help you: check https://github.com/sigrokproject/libsigrok/blob/master/src/libsigrok-internal.h
- Don't use asserts. Check for errors, report, return error code.
- Don't leave TODO comments, #ifdef'd out code, debug output etc. in your code.
- Don't invent your own macros for compiler attributes.
- Don't do big macros and inline functions. Let the compiler decide what to do with your function.
- Don't hardcode blocking serial command and reply functions, this blocks the frontend for no good reason. Just keep state and do what needs doing when you get control after a read, timeout, etc. You're already doing this sort of thing in `peaktech_next_state()`, why include this long blocking read-reply at 9600 bps?
- It's fine to not support every possible untested model or hardware feature, but don't leave TODOs in the code to that effect. Rather make a wiki page for the hardware, and make a section there asking people to help out.
- Is there really no way to detect the model in the protocol? `force_detect` is kind of a last resort. Of course some hardware just leaves you no choice.

ensc:
> * We like drivers to follow a particular pattern -- `api.c` for driver registration and its callbacks, `protocol.c` for hardware-specific code, and `protocol.h` for definitions needed by both

patchset has this split.  `api.c` is for sigrok specific stuff, `protocol.[ch]` contains the protocol specific parts.


> We have a specific coding style in the project, and need contributors to follow it. Even stuff that isn't explicitly spelled out -- lining up variable and structs with tons of whitespace, say

sorry, I do not see the problem...


> there are lots of internal macros to help you: check https://github.com/sigrokproject/libsigrok/blob/master/src/libsigrok-internal.h

functionality there is insufficient and written in a bad style.  E.g. there are only functions which read from `uint8_t *` memory locations.  There is no type safety and you have probably to do manual casts.


> Don't use asserts. Check for errors, report, return error code.

ok


> Don't leave TODO comments, #ifdef'd out code,

what is wrong with "TODO" in comments?  I would say, `#ifdef`ing out dead code is bad.


> Don't invent your own macros for compiler attributes.

can you help me please and tell the existing one for `__attribute__((__packed__))`?


> Don't do big macros and inline functions. Let the compiler decide what to do with your function.

I do not see the problem in my code.  The `inline` are required and I do not see how to solve the problem in C  in a better way without the non-trivial `#define`


> Don't hardcode blocking serial command and reply functions, this blocks the frontend for no good reason. Just keep state and do what needs doing when you get control after a read, timeout, etc. You're already doing this sort of thing in peaktech_next_state(), why include this long blocking read-reply at 9600 bps?

It is a long time since I wrote the patch, but afair the async stuff works only when acquisition is running.  When it is not running, you have to use blocking i/o.


> It's fine to not support every possible untested model or hardware feature, but don't leave TODOs in the code to that effect

why?


> Is there really no way to detect the model in the protocol? force_detect is kind of a last resort.

This should be already done: when `force_detect` is not given, `peaktech_scan_mode()` is run.

biot:
> I do not see the problem in my code.

Ok.

PR #184, opened 2022-05-31
hantek-dso: add "forced" trigger source that captures frames without waiting for trigger

sustmi:
This special trigger source mimics the trigger mode that is called
"Auto" in the original Hantek's "DSO2090 Software" and in
https://github.com/OpenHantek/OpenHantek6022 open-source project.
In the original https://github.com/OpenHantek/openhantek open-source
project the trigger mode is called "Wait/Force".

This feature can be used to immediately capture a single frame.
Example usage with sigrok-cli:
```
sigrok-cli --driver hantek-dso --frames 1 --config triggersource=forced
```

This enables us to use the oscilloscope for data logging in scripts.

Only after I implemented this I realized that if you have an extra probe,
you can set the triggersource to "EXT" and connect the EXT input
to the 1kHz calibration signal output on the back of the device
and that will trigger every millisecond. However, I still think that
the "forced" trigger source can be useful.

### DIFF
--- a/src/hardware/hantek-dso/api.c
+++ b/src/hardware/hantek-dso/api.c
@@ -168,8 +168,7 @@ static const uint64_t vdivs[][2] = {
 };
 
 static const char *trigger_sources[] = {
-	"CH1", "CH2", "EXT",
-	/* TODO: forced */
+	"CH1", "CH2", "EXT", "forced"
 };
 
 static const char *trigger_slopes[] = {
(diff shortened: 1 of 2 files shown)


acassis (approved):

abraxa:
Merged as https://sigrok.org/gitaction?p=libsigrok.git;a=commit;h=c93f113879a099cdb889612d4dbc7f642d88c62a, thank you!

PR #186, opened 2022-06-25
input: add import module for tektronix isf file format

filipkosecek:


acassis:
ping @filipkosecek

src/input/isf.c
```diff
@@ -0,0 +1,529 @@
...
+
+    return data_ptr;
+}
+
+static void extract_channel_name(struct context *inc, const char *beg)
+{
+    size_t i, channel_ix;
+
+    channel_ix = 0;
+    /* Isf wfid looks something like WFID "Ch1, ..."; thus we must skip character '"' */
```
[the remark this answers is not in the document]
filipkosecek: Sorry for being inactive for a long time. This comment was supposed to mean that WFID follows this pattern in ISF format. So Isf was not a typing error, in case I understand your review correctly.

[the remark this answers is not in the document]
fenugrec: suggest writing acronyms  in all-caps ?, "ISF WFID" looks a lot less accidental

fenugrec (commented):
very sparsely commented, so that without a prior knowledge of the format details, it's hard to properly review the implementation.

Instead of merging master back into this branch, you should probably be rebasing on top of master.

disclaimer : I am neither maintainer nor committer in this project.

marcows:
I'm not quite convinced of the ISF file format as it seems possible to differ dependent on the device. However, the import of my .isf example was more than twice as fast as the same data from .csv file and you don't need to configure the import.

There are a few coding style violations which happen throughout the isf.c file, affected rules are:
- use tabs for indentation, not spaces
- add a space between statement and opening parenthesis
- add a space before opening brace in struct declaration
- put the case statement to the same indentation level as its switch counterpart

As was already mentioned, the merge commit doesn't belong here.

src/input/isf.c
```diff
@@ -0,0 +1,529 @@
...
+            else if(strncmp(beg, "MSB", 3) == 0)
+                inc->byte_order = MSB;
+            else
+                return SR_ERR_DATA;
+            
+            break;
+        case BN_FMT:
+            if(strncmp(beg, "RI", 2) == 0)
+                inc->bn_fmt = RI;
+            else if(strncmp(beg, "FP", 2) == 0)
```
marcows, line 224: In [MSO4000 and DPO4000 Series Digital Phosphor Oscilloscopes Programmer Manual](https://download.tek.com/manual/077024801web.pdf) there are `RI` (signed integer) and `RP` (positive integer), see `WFMOutpre:BN_Fmt`.

Where did you get `FP` (floating point)?

marcows (commented):
I didn't make a full review, just focussed on file format specific stuff.

src/input/isf.c
filipkosecek: I used [this manual](https://www.manualslib.com/manual/1375808/Tektronix-Tds5000b-Series.html?page=816) which specifies `FP` (floating point).

src/input/isf.c
```diff
@@ -0,0 +1,796 @@
...
+}
+
+/* Extract enconding type from the header. */
+static int find_encoding(const char *buf, size_t buflen)
+{
+	char value[MAX_ENCODING_STRING_SIZE];
+
+	find_string_value(buf, buflen, value, MAX_ENCODING_STRING_SIZE);
+
+	/* "BIN" and "BINARY" are accepted as suggested in a pull request comment. */
```
fenugrec, line 235: probably not necessary to specify the idea was from a PR comment

fenugrec, line 235: probably not necessary to specify that the idea was from a PR comment

src/input/isf.c
```diff
@@ -0,0 +1,796 @@
...
+
+/* Locate and extract the channel name in the header. */
+static void extract_channel_name(struct context *inc, const char *buf, size_t buflen)
+{
+	size_t i, channel_ix;
+
+	channel_ix = 0;
+	/* ISF WFID looks something like WFID "Ch1, ..."; hence we must skip character '"' */
+	i = 1;
+	while (i < buflen && buf[i] != ',' && buf[i] != '"' && channel_ix < MAX_CHANNEL_NAME_SIZE - 1)
```
fenugrec, line 205: not a fan of this : very long line, with many conditions, and the following instruction block isn't braced. Makes this hard to process if editor wraps the line due to limited width

src/input/isf.c
```diff
@@ -0,0 +1,796 @@
...
+	if (data_ptr == NULL)
+		return NULL;
+
+	data_ptr += strlen(curve);
+	offset = data_ptr - buf->str;
+	if (offset >= buf->len)
+		return NULL;
+
+	/* Curve metadata length is an ASCII byte, hence -48. */
+	metadata_length = (size_t) *data_ptr - 48;
```
fenugrec, line 181: do you mean the length is encoded as an ascii digit '0' to '9' ? or that they simply mapped lengths 0 to (255-48) to chars 0x30 to 0xff ?

Suggest bounds checking before subtracting, in case of malformed packet.

src/input/isf.c
```diff
@@ -0,0 +1,796 @@
...
+
+/*
+ * Check if the format matches ISF format.
+ *
+ * TODO: The header could be searched for more items
+ * aside from NR_PT to increase the confidence.
+ */
+static int format_match(GHashTable *metadata, unsigned int *confidence)
+{
+	const char default_extension[] = ".isf";
```
fenugrec, line 460: suggest to do extension matching with non-case-sensitive functions (some files may end up with uppercase .ISF and then fail on linux/mac)

src/input/isf.c
```diff
@@ -0,0 +1,796 @@
...
+	unsigned int bytnr;
+	int i;
+	int64_t value;
+	uint8_t data[MAX_INT_BYTNR];
+
+	inc = in->priv;
+	bytnr = inc->bytnr;
+
+	/* Value bytnr is checked in function "receive". */
+	memcpy(data, in->buf->str + offset, bytnr);
```
fenugrec, line 520: bounds check on bytnr before memcpy ? not sure if that's what you mean in the comment on the previous line
[EDIT] ok, yes I see now, it is already checked. But I would probably just replace that comment with the bounds check, even if it's redundant - very inexpensive precaution and 1% easier to review

src/input/isf.c
filipkosecek: The length is encoded as an ascii digit '0' to '9'. Fixed the comment and bounds checking is performed.

src/input/isf.c
filipkosecek: I braced the instruction block and wrapped the condition line.

src/input/isf.c
filipkosecek: The extension matching is now done using a non-case-sensitive function.

src/input/isf.c
filipkosecek: Added bounds checking on bytnr.

src/input/isf.c
```diff
@@ -468,22 +474,21 @@ static int format_match(GHashTable *metadata, unsigned int *confidence)
...
 	/* Check if the header contains NR_PT item. */
 	if (buf == NULL || g_strstr_len(buf->str, buf->len, nr_pt) == NULL)
 		return SR_ERR;
 
 	/* The header contains NR_PT item, the confidence is high. */
 	*confidence = 50;
+
+	/* Increase the confidence if the extension is '.isf'. */
+	fn = g_hash_table_lookup(metadata, GINT_TO_POINTER(SR_INPUT_META_FILENAME));
+	if (fn != NULL && (fn_len = strlen(fn)) >= strlen(default_extension)) {
```
fenugrec, line 487: Also, I didn't catch this last time but you have an assignment inside a conditional, which should be split (I think that is now mentioned in the HACKING doc)

src/input/isf.c
marcows: Instead of using `48` you should use `'0'` to make this obvious.

fenugrec: > Instead of using `48` you should use `'0'` to make this obvious.

Agreed. This is something often written `if (val > '9' ||  val < '0') return bad;  metadata_length = val - '0'`

src/input/isf.c
filipkosecek: Fixed.

src/input/isf.c
filipkosecek: Replaced 48 with '0' to make it obvious.

src/input/isf.c
```diff
@@ -287,12 +286,16 @@ static int parse_isf_header(GString *buf, struct context *inc)
...
+	/* Search for all header items. */
 	for (i = 0; i < HEADER_ITEMS_PARAMETERS; ++i) {
 		pattern = find_item(buf, header_items[i]);
-		if (pattern == NULL)
-			/* Ignore unknown command/header.  */
-			continue;
+		if (pattern == NULL) {
+			/* WFID is not required. */
+			if (i == WFID)
+				continue;
```
marcows, line 441: For me (.isf created by Tektronix DPO 4034) "WFMTYPE" was the problem, not "WFID", as could be determined from the .isf header I had posted:
https://github.com/sigrokproject/libsigrok/pull/186#discussion_r1083373879

The change in your previous commit (the removed lines above had been added there), which ignored all unknown header items (proposal from my review), worked for me. After this change it didn't work anymore.

filipkosecek: The thing is that it is pretty hard to identify which header items should be mandatory as the header slightly varies depending on the device used.

Do you think it would be better to initialize header items to default values and make them all optional?

marcows: I'm not sure because I didn't had a deep look at some specifications. But I don't think there are sensible default values possible in most cases.

I couldn't find WFMTYPE in the manual linked in the comment at the beginning of isf.c file.

filipkosecek: The default option for WFMTYPE could be `ANALOG` value. The other options are probably rarely used since they are used for radio frequency data. The options are specified in this [manual](https://download.tek.com/manual/MSO3000-and-DPO3000-Programmer-Manual.pdf) on page 2-586.

PR #190, opened 2022-08-09
Add udev rules for Saleae devices.

phodina:
These PIDs and VIDs are based on udev rules from Logic 2 app.

I can confirm the one for Saleae Logic Pro 8. The others should work as well since it comes from official hardware but I don't have HW to test so I can't add comments to which device they correspond.


ususdei:
FYI: `ATTRS{idVendor}=="21a9", ATTRS{idProduct}=="1004"` is the `Saleae Logic 8`  (without "pro").

Unfortunately it is not supported by sigrok yet.

abraxa:
Merged as https://sigrok.org/gitaction?p=libsigrok.git;a=commit;h=5aada12bfff86a233a82721cebafa53f36e136cc , commenting out the Saleae Pro 8 but leaving it in for reference. Thank you!

PR #191, opened 2022-08-09
korad-kaxxxxp: Added support for TENMA 72-2705 V6.6 (1Ch, 30v, 3A)

dries007:
Tested with hardware.


knarfS:
This PR can be closed, power supply was added with commit 8c08409

PR #194, opened 2022-09-09
agilent-dmm: Fix compatibility with Agilent U1242A

bwalex:
The U1242A sends conf responses as follows,

    CONF? response '"VOLT +1.000000E+00,+1.000000E-04"

    CONF? response '"TEMP:ENV CEL"'

So update the regex patterns to cope with these.


fenugrec:
Looks OK,  but it would be good to have confirmation from other agilent-dmm users to confirm it doesn't break support for other devices.

abraxa:
Merged as https://sigrok.org/gitaction?p=libsigrok.git;a=commit;h=83d14190b95707d7201555122317185520c80bd0 , thank you!

PR #196, opened 2022-09-16
korad-kaxxxxp: Try to identify devices without version string.

duc996:
Devices are identified with the whole *IDN? response (except the serial number). This patch removes the version information check.


fenugrec:
Suggest adding a comment showing different typical responses, with/without serial number - would help analyzing what the code does and should be doing.

duc996:
Thanks for your comment but gsigh has proposed a more elaborated patch which removes also the serial number check (from d8aad72 to 68321f7). So my patch is not needed any more, so I close it.

PR #197, opened 2022-09-27
dmm/bm??x: Add Wayback Machine links to brymen documentation

mmuman:
While checking if a DMM could use an existing driver I found those URLs to be broken.


depili:
The commits should ideally be merged into one and have a better message; something like

> Fix brymen product URL after site restructure
>
> Change http://brymen.com/product-html/ to http://www.brymen.com/

The most ideal thing would be to also include archive.org URLs so they should stay reachable better.

mmuman:
I'll have a look but I'm quite busy this week.

mmuman:
Aw seems someone already merged it manually…

abraxa:
Merged as https://sigrok.org/gitaction?p=libsigrok.git;a=commit;h=70a23b5019ed4d29316683739894fae1abcd7c58 , thank you!

PR #199, opened 2022-11-30
HACKING: add notes for recommended practices

fenugrec:
If these look good, I will prepare a similar update to the other mainline projects.

### DIFF
--- a/HACKING
+++ b/HACKING
@@ -98,6 +98,9 @@ Random notes
    Do use g_try_malloc() or g_try_malloc0() for large (>= 1MB) allocations
    and check the return value.
 
+ - Endianness conversions : prefer the functions provided in
+   libsigrok-internal.h, such as read_be16() etc.
+
  - You should never print any messages (neither to stdout nor stderr nor
    elsewhere) "manually" via e.g. printf() or g_log() or similar functions.
    Only sr_err()/sr_warn()/sr_info()/sr_dbg()/sr_spew() should be used.
--- a/HACKING
+++ b/HACKING
@@ -79,6 +79,9 @@ Random notes
  - Generally avoid assigning values to variables at declaration time,
    especially so for complex and/or run-time dependent values.
 
+ - Separate assignments from control flow. Example : avoid the pattern
+   if (var = func()) {...}  as it complicates review and maintenance.
+
  - Consistently use g_*malloc() / g_*malloc0(). Do not use standard
    malloc()/calloc() if it can be avoided (sometimes other libs such
    as libftdi can return malloc()'d memory, for example).


HACKING
```diff
@@ -95,6 +98,9 @@ Random notes
    Do use g_try_malloc() or g_try_malloc0() for large (>= 1MB) allocations
    and check the return value.
 
+ - Endianness conversions : prefer the functions provided in
+   libsigrok-internal.h, such as read_be16() etc.
```
ensc, line 102: I do not think this should be enforced.  This api is poorly designed and unsafe:

- you have to work with pointers; there is no way to convert a value.   This might be useful when you read a stream, but for packet based communications, this makes the code ugly and error prone.

- related functions take an `uint8_t *` pointer; this requires (unsafe) casts and expressions like `R8(23)` are accepted without any warning.  When you really want to go this way, `void *` should be used.

There are much better ways to deal with endian conversions.

fenugrec: Thanks for your comments. Happy to discuss this, even though my opinion here has little influence.

Do you agree that there are situations where we obtain stuff into a u8 buffer, possibly without the proper alignment, meaning that casting any part of it as anything larger than a u8, (e.g. `struct *` , `uint16 *`) etc is wrong ? For these cases, you need something like `read_be16()` . 

Do you also agree that funcs like `htons` are opaque and not very helpful ? 

> expressions like R8(23) are accepted without any warning

Agreed, I believe that's why gsigh prefers avoiding the Rxx macros, and also why I didn't mention them.


> There are much better ways to deal with endian conversions.

Please, continue.

ensc: > Do you agree that there are situations where we obtain stuff into a u8 buffer

It depends on your code and type of data.  When you have a very stream based protocol or arrays with variable size, a `read_xx()` approach might be useful.

But when you work with something which can be expressed as

```c
struct foo {
        uint8_t a;
        uint32_t b;
        uint8_t c;
} __attribute__((__packed__));
```

you should use this `struct` directly.  Compiler knows how to access the unaligned `b`.

It can be used like [here](https://github.com/sigrokproject/libsigrok/blob/496919362b3dba2f0be340cb1a96dad68879de34/src/hardware/peaktech-607x/protocol.h#L111-L122).

> Do you also agree that funcs like htons are opaque and not very helpful ?

I do not know, what you mean with "opaque".  Problem with `htons` is, that it comes from a completely different area (`#include <arpa/inet.h>`)  and that there is no 64 bit variant.  The `s` sounds also like `short` (but it is `uint16_t`).

The functions from `<endian.h>` are probably a better replacement but are from BSD and not available everywhere.


>> There are much better ways to deal with endian conversions.
> Please, continue.

from my PeakTech 6070 PR:

https://github.com/sigrokproject/libsigrok/blob/496919362b3dba2f0be340cb1a96dad68879de34/src/hardware/peaktech-607x/endian.h
https://github.com/sigrokproject/libsigrok/blob/496919362b3dba2f0be340cb1a96dad68879de34/src/hardware/peaktech-607x/protocol.h

Basically, there are strongly typed

```c
typedef struct be16 {
    uint16_t v;
} be16_t;
```

helpers which can be accessed by special functions (`uint16_t be16_to_cpu(be16_t v)`).    Because these functions take parameters as value (not reference), there are no alignment issues in the functions itself and caller copies the (perhaps unaligned) data as described above.

The strong typing prevents accidental direct access like

```c
be16_t foo = 23;   // error
be16_t foo = cpu_to_be16(23); // ok
```

There should be no runtime impact (recent compilers optimize the `struct` away); but to be sure, you can put this in something like

```c
#ifndef RELEASE
typedef struct be16 { ... } be16_t;

#else
typedef uint16_t be16_t;

#endif
```

fenugrec: @harper23 may be interested in this discussion, related to https://github.com/sigrokproject/libsigrok/commit/c3d4003710edfe193de072f6032401bcba098672

fenugrec:
(force-pushed to reword the commit messages as suggested on the ML)

fenugrec:
merged in 88276d9ddca418c724a7a35719185a92f747659c

PR #201, opened 2022-12-02
input/csv: fix import of analog value 0 in second and subsequent chunks

marcows:
clear_analog_samples() didn't clear, so old buffer content existed.

I noticed it by sporadic peaks in low bus signal (around 0 voltage).

### DIFF
--- a/src/input/csv.c
+++ b/src/input/csv.c
@@ -362,8 +362,6 @@ static int queue_logic_samples(const struct sr_input *in)
 	return SR_OK;
 }
 
-static void set_analog_value(struct context *inc, size_t ch_idx, csv_analog_t value);
-
 static void clear_analog_samples(struct context *inc)
 {
 	size_t idx;
@@ -372,7 +370,7 @@ static void clear_analog_samples(struct context *inc)
 		return;
 	inc->analog_sample_buffer = &inc->analog_datafeed_buffer[inc->analog_datafeed_buf_fill];
 	for (idx = 0; idx < inc->analog_channels; idx++)
-		set_analog_value(inc, idx, 0.0);
+		inc->analog_sample_buffer[idx * inc->analog_datafeed_buf_size] = 0.0;
 }
 
 static void set_analog_value(struct context *inc, size_t ch_idx, csv_analog_t value)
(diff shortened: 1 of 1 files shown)


pkasemir (approved):
This change looks good, but makes me wonder why `set_analog_value()` doesn't allow setting it to 0.0. I would imagine that 0.0 is a perfectly valid value for any analog signal.

fenugrec:
Well,

```
static void set_analog_value(struct context *inc, size_t ch_idx, csv_analog_t value)
{
        if (ch_idx >= inc->analog_channels)
                return;
        if (!value)
                return;
  ....
```

why is it checking for (!value) ? It's not like it's a pointer, and that kind of check is always suspicious with float / double values

marcows:
In the logic counterpart it is similar, logic zero is not set, only logic one:

```
static void set_logic_level(struct context *inc, size_t ch_idx, int on)
{
	[...]

	if (ch_idx >= inc->logic_channels)
		return;
	if (!on)
		return;

	[...]
```

I can only guess of performance improvements as a reason.

See
```
memset(inc->sample_buffer, 0, inc->sample_unit_size);
```
for clearing all samples/channels of one point in time vs.
```
byte_idx = ch_idx / 8;
bit_idx = ch_idx % 8;
bit_mask = 1 << bit_idx;
inc->sample_buffer[byte_idx] |= bit_mask;
```
for setting one sample/channel.

The samples are always cleared before being set, so setting again can be avoided.

Originally there was only support for logic signals, analog signals have been added later and adopted this approach. But the performance gain (if existing for the logic samples) does not exist there.

marcows:
To reproduce use the attached file [example.csv](https://github.com/sigrokproject/libsigrok/files/10144934/example.csv) like this:
`$ sigrok-cli -D -i example.csv -I csv:column_formats=a | tail`

Expected output:
```
CH1: 5.000 
CH1: 0.000 
CH1: 0.000 
CH1: 0.000 
CH1: 0.000 
CH1: 0.000 
CH1: 0.000 
CH1: 5.000 
CH1: 5.000 
CH1: 5.000
```

We get instead:
```
CH1: 5.000 
CH1: 0.000 
CH1: 0.000 
CH1: 0.000 
CH1: 5.100 
CH1: 5.200 
CH1: 5.300 
CH1: 5.000 
CH1: 5.000 
CH1: 5.000
```

Note: You have to use the current Git state of libsigrok, with the released libsigrok 0.5.2 this bug needs twice as many samples to become effective.

depili:
Merged as https://sigrok.org/gitaction/libsigrok.git/8c08409..e7ed370

Thank you.

marcows:
Implemented differently in commit e7ed37042611345fdcd47dbd6b9e3ec6fc38919a

PR #202, opened 2022-12-11
Fix exceptions with u32 command line options because of incomplete bindings

marcows:
Python/Ruby bindings not tested, only C++ bindings with PulseView.

No fallback implementation of stoul() for Android added. Is this still needed or do current/supported versions provide it? And I didn't want to continue before checking whether I'm on the right way at all.


marcows:
Applied in these commits:
77ee3168d0ae88c92a127999d890fff0511ca31e
f806de1864167feed30952f1ab65497b8ee4152a
0db1b189bee3ffe5c6ea39d7ca2e62715856b538

PR #118, opened 2021-02-06
siglent-sds: Reimplement eseries acquisition logic and add new scope models

voneiden:
Changes are implemented so that they should not affect other series besides eseries. These changes have been using SDS-1104X-E. Fixes broken support to all e-series scopes and adds new models: SDS1104X-U, SDS2202X-E, SDS2352X-E

The acquisition logic for the eseries has been completely reinvented, simplified and made quite error resistant. USBTMC support is included (although it is significantly slower than TCP). The acquisition logic for the eseries is also now completely isolated from older models, allowing maintenance and improvements to be made and tested without breaking legacy compatibility.

Addresses bugs:

* https://sigrok.org/bugzilla/show_bug.cgi?id=1355 (eseries)
* https://sigrok.org/bugzilla/show_bug.cgi?id=1356 (eseries)
* https://sigrok.org/bugzilla/show_bug.cgi?id=1242 (eseries)
* https://sigrok.org/bugzilla/show_bug.cgi?id=1628 (usbtmc/eseries)

PR #203, opened 2022-12-15
siglent-sds: add support for SDS1104X-U

torkildr:
The SDS1104X-U is basically a flavor of the SDS1104X-E without digital channels.


torkildr:
It seems to work well as a `SDS1000XE`. There _might_ be some corner cases where it will behave differently, but I think this is the best starting point

torkildr:
Closing this, as it's a badly duplicated version of #118

PR #205, opened 2023-01-16
Add support for RD6006P and RD6012P

mdjurfeldt:
This PR adds basic support for the models RD6006P and RD6012P. In order to get the current readings correct for RD6012P, I had to dynamically adapt the current_multiplier to which current range (6V/12V) is set for the PSU. In the future, one could also add the possibility to switch voltage range to the GUI.

This has been tested on an RD6012P but should work on all the other models.

### DIFF
--- a/src/hardware/rdtech-dps/protocol.c
+++ b/src/hardware/rdtech-dps/protocol.c
@@ -21,10 +21,12 @@
 
 #include <config.h>
 
+#include <math.h>
 #include <string.h>
 
 #include "protocol.h"
 
+/* These are the Modbus RTU registers for the family of rdtech-dps devices. */
 enum rdtech_dps_register {
 	REG_DPS_USET       = 0x00, /* Mirror of 0x50 */
 	REG_DPS_ISET       = 0x01, /* Mirror of 0x51 */
@@ -69,6 +71,10 @@ enum rdtech_dps_regulation_mode {
 	MODE_CC      = 1,
 };
 
+/*
+ * Some registers are specific to a certain device. For example,
+ * REG_RD_RANGE is specific to RD6012P.
+ */
 enum rdtech_rd_register {
 	REG_RD_MODEL = 0, /* u16 */
 	REG_RD_SERIAL = 1, /* u32 */
@@ -85,6 +91,8 @@ enum rdtech_rd_register {
 	REG_RD_PROTECT = 16, /* u16 */
 	REG_RD_REGULATION = 17, /* u16 */
 	REG_RD_ENABLE = 18, /* u16 */
+	REG_RD_PRESET = 19, /* u16 */
+	REG_RD_RANGE = 20, /* u16 */
 	/*
 	 * Battery at 32 == 0x20 pp:
 	 * Mode, voltage, temperature, capacity, energy.
@@ -108,12 +116,12 @@ static int rdtech_dps_read_holding_registers(struct sr_modbus_dev_inst *modbus,
 	int ret;
 
 	retries = 3;
-	while (retries--) {
+	do {
 		ret = sr_modbus_read_holding_registers(modbus,
 			address, nb_registers, registers);
 		if (ret == SR_OK)
 			return ret;
-	}
+	} while (--retries);
 
 	return ret;
 }
@@ -200,7 +208,7 @@ SR_PRIV int rdtech_dps_get_model_version(struct sr_modbus_dev_inst *modbus,
 		if (ret != SR_OK)
 			return ret;
 		rdptr = (void *)registers;
-		*model = read_u16be_inc(&rdptr) / 10;
+		*model = read_u16be_inc(&rdptr);
 		*serno = read_u32be_inc(&rdptr);
 		*version = read_u16be_inc(&rdptr);
 		sr_info("RDTech RD model: %u version: %u, serno %u",
@@ -211,6 +219,48 @@ SR_PRIV int rdtech_dps_get_model_version(struct sr_modbus_dev_inst *modbus,
 		return SR_ERR_ARG;
 	}
 	/* UNREACH */
+}
+
+SR_PRIV void rdtech_dps_update_multipliers(const struct sr_dev_inst *sdi)
+{
+	struct dev_context *devc;
+	struct rdtech_dps_range *range;
+
+	devc = sdi->priv;
+	range = &devc->model->ranges[devc->curr_range];
+	devc->current_multiplier = pow(10.0, range->current_digits);
+	devc->voltage_multiplier = pow(10.0, range->voltage_digits);
+}
+
+/*
+ * Determine range of connected device. Don't do anything once
+ * acquisition has started (since the range will then be tracked).
+ */
+SR_PRIV int rdtech_dps_update_range(const struct sr_dev_inst *sdi)
+{
+	struct dev_context *devc;
+	uint16_t range;
+	int ret;
+
+	devc = sdi->priv;
+
+	/*
+	 * Only update range if there are multiple ranges and data
+	 * acquisition hasn't started.
+	 */
+	if (devc->model->n_ranges == 1 || devc->acquisition_started)
+		return SR_OK;
+	if (devc->model->model_type != MODEL_RD)
+		return SR_ERR;
+
+	ret = rdtech_dps_read_holding_registers(sdi->conn,
+		REG_RD_RANGE, 1, &range);
+	if (ret != SR_OK)
+		return ret;
+	devc->curr_range = range ? 1 : 0;
+	rdtech_dps_update_multipliers(sdi);
+
+	return SR_OK;
 }
 
 /* Send a measured value to the session feed. */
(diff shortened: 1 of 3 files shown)


src/hardware/rdtech-dps/protocol.c
```diff
@@ -200,7 +202,7 @@ SR_PRIV int rdtech_dps_get_model_version(struct sr_modbus_dev_inst *modbus,
 		if (ret != SR_OK)
 			return ret;
 		rdptr = (void *)registers;
-		*model = read_u16be_inc(&rdptr) / 10;
+		*model = read_u16be_inc(&rdptr);
```
fenugrec, line 205: can you elaborate on why this was dividing the model number by ten a model number, and why it's not necessary anymore ? Does that risk breaking anything ?

src/hardware/rdtech-dps/protocol.c
```diff
@@ -85,6 +85,8 @@ enum rdtech_rd_register {
 	REG_RD_PROTECT = 16, /* u16 */
 	REG_RD_REGULATION = 17, /* u16 */
 	REG_RD_ENABLE = 18, /* u16 */
+	REG_RD_PRESET = 19, /* u16 */
+	REG_RD_RANGE = 20, /* u16 */
```
fenugrec, line 89: I'd like a note that says whether these numeric values are arbitrary or part of the protocol. In contrast to the 'state mask' enum that appears to be arbitrary and only for maintaining local state ?

src/hardware/rdtech-dps/protocol.c
```diff
@@ -396,6 +409,10 @@ SR_PRIV int rdtech_dps_get_state(const struct sr_dev_inst *sdi,
 		is_reg_cc = reg_state == MODE_CC;
 		out_state = read_u16be_inc(&rdptr); /* ENABLE */
 		is_out_enabled = out_state != 0;
+		if (devc->model->id == 60125) {
```
fenugrec, line 412: I think magic number  "60125" deserves a macro name....

fenugrec (commented):
Also I think the commit message's first line should be changed to something like "rdtech-dps: <change summary>" - see other commits in the repo

src/hardware/rdtech-dps/protocol.c
mdjurfeldt: Thank you for your question.

The old code used this kludge (dividing by 10) to extract the textual model designation from the actual numerical model number. This worked as long as there were numerical model numbers such as 60061, 60121, 60241. However the models 6006P and 6012P have the numerical model numbers 60065 and 60125. Dividing these by 10 throws away information such that one can no longer distinguish between e.g. 6006 and 6006P. In my mind, it is better to store the model number chosen by the manufacturer than to come up with an alternative scheme, especially if information is lost. This stored model number is anyway used to lookup the textual model designation from a table before displaying it in the GUI. This is the only use of the numerical model number in the old code that I'm aware of. In my PR, I, in addition, use the numerical model number to conditionally run some code.

So, no, I don't think this breaks anything. In fact, it makes the code more future safe, if Ruideng comes up with another new series of devices.

src/hardware/rdtech-dps/protocol.c
mdjurfeldt: They are not arbitrary. These are the device family specific modbus registers. I think that you have a point that this should be stated in a comment. I'll do that.

src/hardware/rdtech-dps/protocol.c
mdjurfeldt: Well, in the general case I of course agree that there should not be magic numbers in the code. The reason why I did it this way in this case is that 60125 is not magic but is the most specific name of the model. Thus, a suitable macro name could be MODEL_NUMBER_60125 which is kind of pointless.

What would you say about defining a macro (in protocol.c) containing 60125 but used in this location as:

  `if (IS_MODEL_6012P (devc->model))`

?

mdjurfeldt:
@fenugrec Thank you for your review.

I have responded to your questions and have pushed new changes that address the issues you raised. I had to force push to modify the commit messages to the rdtech-dps: style.

src/hardware/rdtech-dps/protocol.c
fenugrec: Ok, fair. Was not aware of what the "model string" actually looked like.

mdjurfeldt:
Hi @fenugrec, are you OK with the current state of this PR? Meanwhile, I've learnt that the preferred way to report the review is not github but the mailinglist sigrok-devel@lists.sourceforge.net. Can I ask you to send an email to that list saying that you have reviewed this change?

fenugrec:
> Can I ask you to send an email to that list saying that you have reviewed this change?

No time for that. You can post yourself, AFAIK registering should be quick

Sidenote, now you will need much patience - sigrok is beyond understaffed.

mdjurfeldt:
OK, done! Many thanks for your review. I'll keep your suggestion to use the table for feature flags in mind for the future.

src/hardware/rdtech-dps/protocol.c
```diff
@@ -105,7 +114,7 @@ static int rdtech_dps_read_holding_registers(struct sr_modbus_dev_inst *modbus,
 	int address, int nb_registers, uint16_t *registers)
 {
 	size_t retries;
-	int ret;
+	int ret = SR_ERR; /* make the compiler happy */
```
knarfS, line 117: Only variable declaration here

src/hardware/rdtech-dps/protocol.c
```diff
@@ -105,7 +114,7 @@ static int rdtech_dps_read_holding_registers(struct sr_modbus_dev_inst *modbus,
...
 
 	retries = 3;
```
knarfS, line 119: Variable assignment `ret = SR_ERR` here

src/hardware/rdtech-dps/protocol.c
```diff
@@ -259,13 +277,13 @@ SR_PRIV int rdtech_dps_get_state(const struct sr_dev_inst *sdi,
...
 	gboolean get_config, get_init_state, get_curr_meas;
-	uint16_t registers[12];
+	uint16_t registers[14];
 	int ret;
 	const uint8_t *rdptr;
 	uint16_t uset_raw, iset_raw, uout_raw, iout_raw, power_raw;
 	uint16_t reg_val, reg_state, out_state, ovpset_raw, ocpset_raw;
 	gboolean is_lock, is_out_enabled, is_reg_cc;
-	gboolean uses_ovp, uses_ocp;
+	gboolean uses_ovp, uses_ocp, uses_12V_range = FALSE;
```
knarfS, line 286: See above: variable declaration / assignment

knarfS (changes requested):

src/hardware/rdtech-dps/protocol.c
mdjurfeldt: Fixed in commit b7672857

mdjurfeldt:
@knarfS @fenugrec: Hi, I've enabled listing, setting and getting range now. I also hope that everything is sync:ed with the device now and that range meta packages are sent if range is changed.

mdjurfeldt:
@knarfS @fenugrec: Maybe I have been overambitious regarding the updating of range and multipliers? I have operated under the assumption that range can change by manipulating the device manually during sigrok control. However, when sigrok opens the device it actually locks it. So, maybe we *can* make some assumptions regarding when range can change (and, thus, simplify the code and avoid some communication)?

src/hardware/rdtech-dps/protocol.c
```diff
@@ -560,6 +620,32 @@ SR_PRIV int rdtech_dps_set_state(const struct sr_dev_inst *sdi,
...
+				/* No need to set. */
+				return SR_OK;
+			ret = rdtech_rd_set_reg(sdi, REG_RD_RANGE, reg_value);
+			if (ret != SR_OK)
+				return ret;
+			if (!devc->acquisition_started) {
+				devc->curr_range = reg_value ? 1 : 0;
+				rdtech_dps_update_multipliers(sdi);
+			}
+			/* We rely on the data acquisition to update
```
knarfS, line 641: See style nit for comments above

src/hardware/rdtech-dps/protocol.h
```diff
@@ -69,18 +77,21 @@ struct rdtech_dps_state {
...
 		STATE_PROTECT_OCP = 1 << 4,
-		STATE_PROTECT_ENABLED = 1 << 5,
-		STATE_VOLTAGE_TARGET = 1 << 6,
-		STATE_CURRENT_LIMIT = 1 << 7,
-		STATE_OVP_THRESHOLD = 1 << 8,
-		STATE_OCP_THRESHOLD = 1 << 9,
-		STATE_VOLTAGE = 1 << 10,
-		STATE_CURRENT = 1 << 11,
-		STATE_POWER = 1 << 12,
+		STATE_RANGE = 1 << 5,
```
knarfS, line 80: Better append the new status at the end.

knarfS (changes requested):

src/hardware/rdtech-dps/protocol.c
```diff
@@ -658,6 +749,13 @@ SR_PRIV int rdtech_dps_receive_data(int fd, int revents, void *cb_data)
 			g_variant_new_boolean(state.protect_ocp));
 		devc->curr_ocp_state = state.protect_ocp;
 	}
+	if (devc->curr_range != state.range) {
+		(void)sr_session_send_meta(sdi,
+			SR_CONF_RANGE,
+			g_variant_new_uint32(state.range));
```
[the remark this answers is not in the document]
mdjurfeldt: Ah... me stupid.

src/hardware/rdtech-dps/protocol.c
```diff
@@ -369,7 +419,9 @@ SR_PRIV int rdtech_dps_get_state(const struct sr_dev_inst *sdi,
 		/* Retrieve a set of adjacent registers. */
 		g_mutex_lock(&devc->rw_mutex);
 		ret = rdtech_dps_read_holding_registers(modbus,
-			REG_RD_VOLT_TGT, 11, registers);
+			REG_RD_VOLT_TGT,
+			devc->model->n_ranges > 1 ? 13 : 11,
```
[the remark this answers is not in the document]
mdjurfeldt: OK, I'll do this in protocol.c since the registers are listed there.

knarfS (approved):

abraxa:
Merged as https://sigrok.org/gitaction/libsigrok.git/48385ca..02a4f48 , thank you!

Also lots of thanks for @fenugrec and @knarfS for reviewing, made my life much easier!

PR #206, opened 2023-01-21
Agilent 54621d device driver

taragor:
This is my implementation of a device driver for the Agilent 54621d mixed signal oscilloscope. 
The driver allows to configure the device and download data from memory or single shot mode.

Downloading data is done in a hacky way: 
The device natively only allows downloading either upto 2k Points from the display buffer or the entire captured waveform (upto 8M points). Downloading the entire waveform takes ages (~10Minutes per channel) since the Device only allows upto 57600Baud. To speed up download time I've implemented Downloading so it zooms and pans the captured waveform to use the Scopes display decimation filter to downsample to the requested sample rate and then download chunks of 2k points until the requested number of points is downloaded.

I'm planning to implement high resolution mode and pattern trigger, but other than that the driver is finished from my point of view.

It can be generalized to work with all devices of the 546xx family, however I've not yet gotten around to doing that and cannot test the other devices. (Might be able to get access to a 54624a)

### DIFF
--- a/src/hardware/agilent-54621d/api.c
+++ b/src/hardware/agilent-54621d/api.c
@@ -1,7 +1,7 @@
 /*
  * This file is part of the libsigrok project.
  *
- * Copyright (C) 2022 Daniel <1824222@stud.hs-mannheim.de>
+ * Copyright (C) 2022 Daniel Echt <taragor83@gmail.com>
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -60,30 +60,6 @@ static const uint64_t samplerates[] = {
 	SR_HZ(1),
 };
 
-static const uint64_t samplerates_literal[] = {
-	SR_KHZ(2),
-	SR_KHZ(4),
-	SR_KHZ(5),
-	SR_KHZ(10),
-	SR_KHZ(20),
-	SR_KHZ(40),
-	SR_KHZ(50),
-	SR_KHZ(100),
-	SR_KHZ(200),
-	SR_KHZ(400),
-	SR_KHZ(500),
-	SR_MHZ(1),
-	SR_MHZ(2),
-	SR_MHZ(4),
-	SR_MHZ(5),
-	SR_MHZ(10),
-	SR_MHZ(20),
-	SR_MHZ(25),
-	SR_MHZ(50),  //Can appear on device
-	SR_MHZ(100), //Can appear on device
-	SR_MHZ(200), //Can appear on device
-};
-
 static struct sr_dev_inst *probe_device(struct sr_scpi_dev_inst *scpi)
 {
 	struct sr_dev_inst *sdi;
@@ -135,6 +111,19 @@ static GSList *scan(struct sr_dev_driver *di, GSList *options)
 {
 	sr_info("Scanning for agilent 54621d");
 	return sr_scpi_scan(di->context, options, probe_device);
+}
+
+static void clear_helper(struct dev_context *devc)
+{
+	agilent_54621d_scope_state_free(devc->model_state);
+	g_free(devc->analog_groups);
+	g_free(devc->digital_groups);
+	g_free(devc->data);
+}
+
+static int dev_clear(const struct sr_dev_driver *di)
+{
+	return std_dev_clear_with_callback(di, (std_dev_clear_callback)clear_helper);
 }
 
 static int dev_open(struct sr_dev_inst *sdi)
@@ -652,9 +641,7 @@ static int dev_acquisition_start(const struct sr_dev_inst *sdi)
 	struct sr_scpi_dev_inst *scpi;
 	struct scope_state *state;
 	const struct scope_config *model;
-	int i;
 	int tmp_int;
-	gboolean acq_complete;
 	gboolean tmp_bool;
 	char command[MAX_COMMAND_SIZE];
 	float xinc;
@@ -664,9 +651,6 @@ static int dev_acquisition_start(const struct sr_dev_inst *sdi)
 	struct analog_channel_transfer_info *encoding;
 	gboolean digital_added[MAX_DIGITAL_GROUP_COUNT];
 	size_t group, pod_count;
-	struct sr_datafeed_packet packet;
-	struct sr_datafeed_header header;
-	struct timeval timeToTrigger;
 
 
 	scpi = sdi->conn;
@@ -720,7 +704,6 @@ static int dev_acquisition_start(const struct sr_dev_inst *sdi)
 	if(devc->data_source == DATA_SOURCE_LIVE){
 		sr_scpi_get_bool(scpi, ":TER?", &tmp_bool);
 		sr_scpi_send(scpi, ":SING");
-		acq_complete = FALSE;
 		sr_scpi_get_bool(scpi, ":TER?", &tmp_bool);
 		while(!tmp_bool){
 			g_usleep(WAIT_FOR_CAPTURE_COMPLETE_DELAY);
@@ -747,13 +730,12 @@ static int dev_acquisition_start(const struct sr_dev_inst *sdi)
 	* The logic is, that if we want less than the complete waveform we can switch to window view, calculate the settings to have the window show exactly 2k points of the wv, and then transfer these 2k points.
 	* Then we can move the window delay and download the next 2k points. We can repeat this until we have the desired amount of points. 
 	* This is also how the scope transfers the full waveform, however transfering the full wavefrom cannot be interrupted, so the manual approach is better
-	* We save the number of samples already downloaded in devc->num_samples, and the samples themselves in devc->buffer until all data for a channel is ready
 	*/
 
 	//Reset parameters
 	devc->num_samples = 0;
 	devc->num_blocks_downloaded = 0;
-	devc->buffer = g_malloc0(sizeof(char) * devc->samples_limit);
+	//devc->buffer = g_malloc0(sizeof(char) * devc->samples_limit);
 	devc->headerSent = FALSE;
 
 	//set waveform source channel to the first channel to be downloaded
@@ -812,7 +794,7 @@ static int dev_acquisition_start(const struct sr_dev_inst *sdi)
 	g_snprintf(command, sizeof(command), ":TIM:MODE MAIN;:TIM:RANG %f;:TIM:DEL %f", devc->block_deltaT, devc->timebaseLbound-devc->refPos*devc->block_deltaT*0.1);
 	sr_scpi_send(scpi, command);
 
-	sr_dbg("Download %d packets with a width of %f. Maxpoints are %d. Lbound is %f. Trigger at sample %d", devc->num_block_to_download, devc->block_deltaT, points, devc->timebaseLbound, devc->trigger_at_sample);
+	sr_dbg("Download %d packets with a width of %f. Maxpoints are %d. Lbound is %f. Trigger at sample %ld", devc->num_block_to_download, devc->block_deltaT, points, devc->timebaseLbound, devc->trigger_at_sample);
 
 	sr_dbg("beginning data download");
 	//make final setup before download
@@ -897,6 +879,7 @@ static int dev_acquisition_stop(struct sr_dev_inst *sdi)
 	const struct scope_config *model;
 	struct scope_state *state;
 	float timebase;
+	struct sr_channel *ch;
 
 	devc = sdi->priv;
 	model = devc->model_config;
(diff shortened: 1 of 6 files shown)


src/hardware/agilent-54621d/api.c
```diff
@@ -0,0 +1,927 @@
...
+	* Downloading data is done in a quite manual way to speed up downloading. The device only allows downloading upto 2k points from the view-buffer, or the complete captured waveform (~1m points)
+	* The logic is, that if we want less than the complete waveform we can switch to window view, calculate the settings to have the window show exactly 2k points of the wv, and then transfer these 2k points.
+	* Then we can move the window delay and download the next 2k points. We can repeat this until we have the desired amount of points. 
+	* This is also how the scope transfers the full waveform, however transfering the full wavefrom cannot be interrupted, so the manual approach is better
+	*/
+
+	//Reset parameters
+	devc->num_samples = 0;
+	devc->num_blocks_downloaded = 0;
+	//devc->buffer = g_malloc0(sizeof(char) * devc->samples_limit);
```
fenugrec, line 738: dead code

src/hardware/agilent-54621d/api.c
```diff
@@ -0,0 +1,927 @@
...
+	devc->headerSent = FALSE;
+
+	//set waveform source channel to the first channel to be downloaded
+	ch = (struct sr_channel *)devc->enabled_channels->data;
+	if(ch->type == SR_CHANNEL_LOGIC){
+		group = ch->index/DIGITAL_CHANNELS_PER_POD+1;
```
fenugrec, line 744: here and in general : spaces around operators would help readability.

src/hardware/agilent-54621d/api.c
```diff
@@ -0,0 +1,927 @@
...
+		g_snprintf(command, sizeof(command), ":WAV:SOUR POD%ld", group);
+	} else {
+		g_snprintf(command, sizeof(command), ":WAV:SOUR %s", ch->name);	
+	}
+	sr_scpi_send(scpi, command);
+
+	//get required data for calculations
+	sr_scpi_send(scpi, ":TIM:MODE MAIN;:WAV:POIN MAX");
+	sr_scpi_send(scpi, ":WAV:UNS 0");
+	g_usleep(300000);											//Apparently the device needs a little time to do the wav:sour and tim:mode setup before requesting wav:poin? max. waiting 30ms should be enough
```
fenugrec, line 754: magic number. Also 30ms != 300 000 us if I'm mathing right

src/hardware/agilent-54621d/api.c
```diff
@@ -0,0 +1,927 @@
...
+
+	sr_dbg("beginning data download");
+	//make final setup before download
+	sr_scpi_send(scpi, ":WAV:FORM BYTE;BYT MSBF;UNS 0;POIN 2000");	//ToDo: when downloading high resolution data format needs to be word.
+
+	//get header data and store it in channel->priv so the receive_data function can submit it. This only needs to be done for analog channels
+	while(devc->current_channel){
+		ch = (struct sr_channel *)devc->current_channel->data; 
+		if(ch->type == SR_CHANNEL_ANALOG){
+			ch->priv = (void *)g_malloc0(sizeof(struct analog_channel_transfer_info));
```
fenugrec, line 807: useless cast to void *

src/hardware/agilent-54621d/api.c
```diff
@@ -0,0 +1,927 @@
...
+	}
+	sr_scpi_send(scpi, command);
+
+	//register callback
+	sr_scpi_source_add(sdi->session, scpi, G_IO_IN, 100, agilent_54621d_receive_data, (void *)sdi);
+	
+	
+	std_session_send_df_header(sdi);
+
+	//std_session_send_df_header(sdi);
```
fenugrec, line 862: duplicate deadcode

src/hardware/agilent-54621d/api.c
```diff
@@ -0,0 +1,927 @@
...
+	model = devc->model_config;
+	state = devc->model_state;
+
+	std_session_send_df_end(sdi);
+
+	timebase = (float) (*model->timebases)[state->timebase][0] / (*model->timebases)[state->timebase][1];
+
+	while(devc->current_channel){
+		ch = (struct sr_channel *)devc->current_channel->data; 
+			if(ch->type == SR_CHANNEL_ANALOG)
```
fenugrec, line 894: indent is weird in here

src/hardware/agilent-54621d/protocol.c
```diff
@@ -0,0 +1,1032 @@
...
+};
+
+static const char *coupling_options[] = {
+	"AC",
+	"DC",
+	"GND",
+	//Agilent 5464x Scopes also allow 50Ohm Termination, however this is configured using the termination command. If someone wants to implement support for that you need to figure out the best way to do that 
+};
+
+static const char *scope_trigger_slopes[] = {
```
fenugrec, line 122: for this and others where order matters, should have a comment stating so (I think you did it in one place, but these look like they may need it too)

src/hardware/agilent-54621d/protocol.c
```diff
@@ -0,0 +1,1032 @@
...
+	"LIN",
+	"SEQ",
+	"SPI",
+	"TV",
+	"USB",
+};
+
+static const uint64_t scope_timebases[][2] = {
+	/* nanoseconds */
+	{ 5, 1000000000 },
```
fenugrec, line 157: use `1E9` notation please. Counting this many zeros hurts

src/hardware/agilent-54621d/protocol.c
```diff
@@ -0,0 +1,1032 @@
...
+		if(model_index != -1)
+			break;
+	}
+
+	if(model_index == -1){
+		sr_dbg("Unsupported device.");
+		return SR_ERR_NA;
+	}
+
+	devc->analog_groups = g_malloc0(sizeof(struct sr_channel_group*) *scope_models[model_index].analog_channels);
```
fenugrec, line 322: current project guidelines say

> We assume that "small" memory allocations (< 1MB) will always succeed.
   Thus, it's fine to use g_malloc() or g_malloc0() for allocations of
   simple/small structs and such (instead of using g_try_malloc()), and
   there's no need to check the return value.

src/hardware/agilent-54621d/protocol.c
```diff
@@ -0,0 +1,1032 @@
...
+		ch = sr_channel_new(sdi, i, SR_CHANNEL_LOGIC, TRUE,
+			   (*scope_models[model_index].digital_names)[i]);
+
+		group = i / DIGITAL_CHANNELS_PER_POD;
+		devc->digital_groups[group]->channels = g_slist_append(
+			devc->digital_groups[group]->channels, ch);
+	}
+
+	devc->model_config = &scope_models[model_index];
+	devc->samples_limit = 2000;
```
fenugrec, line 364: I think `2000` appears often enough to deserve its own macro

src/hardware/agilent-54621d/protocol.c
```diff
@@ -0,0 +1,1032 @@
...
+	}
+	g_free(tmp_str);
+
+	state->timebase = i;
+
+	//get trigger horizontal position
+	if(sr_scpi_get_float(sdi->conn, (*config->scpi_dialect)[SCPI_CMD_GET_HORIZ_TRIGGERPOS], &tmp_float) != SR_OK)
+		return SR_ERR;
+	
+	state->horiz_triggerpos = tmp_float /
```
fenugrec, line 442: using a float but casting to double...

src/hardware/agilent-54621d/protocol.c
```diff
@@ -0,0 +1,1032 @@
...
+		(((double) (*config->timebases)[state->timebase][0] /
+		  (*config->timebases)[state->timebase][1]) * config->num_xdivs);
+	state->horiz_triggerpos -= 0.5;
+	state->horiz_triggerpos *= -1;
+	//ToDo: This might be dependent on time ref setting
+
+	if (scope_state_get_array_option(sdi->conn,	(*config->scpi_dialect)[SCPI_CMD_GET_TRIGGER_SOURCE], config->trigger_sources, config->num_trigger_sources, &state->trigger_source) != SR_OK)
```
fenugrec, line 449: whitespace offence + long line

src/hardware/agilent-54621d/protocol.c
```diff
@@ -0,0 +1,1032 @@
...
+
+static int scope_state_get_array_option(struct sr_scpi_dev_inst *scpi, const char *command, const char *(*array)[], unsigned int n, int *result)
+{
+	char *tmp;
+	int idx;
+
+	if (sr_scpi_get_string(scpi, command, &tmp) != SR_OK)
+		return SR_ERR;
+
+	if ((idx = std_str_idx_s(tmp, *array, n)) < 0) {
```
fenugrec, line 496: splitting assignment out of conditional would also allow you to `g_free(tmp)` immediately after, in one place only

src/hardware/agilent-54621d/protocol.c
```diff
@@ -0,0 +1,1032 @@
...
+	logic_data = store->data;
+	logic_data += group;
+	logic_step = devc->pod_count;
+	for (idx = 0; idx < pod_data->len; idx++) {
+		*logic_data = pod_data->data[idx];
+		logic_data += logic_step;
+	}
+
+	/* Truncate acquisition if a smaller number of samples has been requested. */
+	if (devc->samples_limit > 0 && devc->logic_data->len > devc->samples_limit * devc->pod_count)
```
fenugrec, line 773: I know this is technically non-ambiguous but can we have a few extra () please

src/hardware/agilent-54621d/protocol.c
```diff
@@ -0,0 +1,1032 @@
...
+
+	if (!(devc = sdi->priv))
+		return TRUE;
+
+	ch = devc->current_channel->data;
+
+	switch (ch->type){
+		case SR_CHANNEL_ANALOG:
+			info = (struct analog_channel_transfer_info *)ch->priv;
+			data = NULL;
```
fenugrec, line 871: data=NULL could go outside the switch{} block instead of each `case` probably

fenugrec (commented):
- didn't do an exhaustive check of indent and whitespace
- didn't ensure that all g_str* operations have a matching free (suggest running valgrind)
- inconsistent braces and newlines with some `if/else` and `} else if ... {`; sometimes should be on the same line. didn't comment inline
- suggest running cppcheck
- some float / double inconsistencies, only commented on some. Suggest looking closely at each explicit cast, g_str functions taking a Double argument, and "%E" formatters
- a few excessively long lines
- a number of assignments-in-conditional
- lacking whitespace around math operators, and patterns like `if(...){`
- Lazy brain so I paid less attention towards the end. Probably missed things.
- `grep "2000" *` should return only a single match, that number appears way too often
- some commented dead code should be removed
- some commits should probably be squashed, since this is an initial implementation ? to check with maintainer
- commit messages should be reworded to follow general format (see log history)

disclaimer : I am neither maintainer nor dev of this project, and there's no guarantee that my suggestions will accelerate processing of this PR.

taragor:
Thanks alot @fenugrec . I'll go over the code and implement the changes. 
How would I go about reducing the commits? Sorry if this is a stupit question, I'm kinda new to this.

src/hardware/agilent-54621d/protocol.c
taragor: This should work just fine when swapped. The index to the currently slected is saved and you actually send "POS" and "NEG" to the device. Since "POS" and "NEG" are also shown in PV it is just looked up whenever it is needed.
Should I add another abstraction to show something else in PV?

src/hardware/agilent-54621d/api.c
```diff
@@ -0,0 +1,927 @@
...
+		}	
+	}
+	devc->current_channel = devc->enabled_channels;
+	devc->pod_count = pod_count;
+
+
+	if(devc->data_source == DATA_SOURCE_LIVE){
+		sr_scpi_get_bool(scpi, ":TER?", &tmp_bool);
+		sr_scpi_send(scpi, ":SING");
+		sr_scpi_get_bool(scpi, ":TER?", &tmp_bool);
```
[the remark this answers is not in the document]
taragor: The first query resets the "Trigger complete" bit in the devices status register. It is not really of interest if it is set before the next acquisition is stated, however I do sr_scpi_get_bool instead of sr_scpi_send have this be blocking. The agilent scope acts somewhat inconsitently if you continue sending commands while it is still sending the reply to a previous query

knarfS:
Here is a good explanation how to squash your commits: https://stackoverflow.com/questions/5189560/how-do-i-squash-my-last-n-commits-together

But keep in mind that your first commit is just the driver skeleton and the second commit is your implementation. Have a look at the [HACKING file in the root of the repository](https://github.com/sigrokproject/libsigrok/blob/master/HACKING).

abraxa:
> It can be generalized to work with all devices of the 546xx family

Would be good if the driver was called 546xx then, too, so that future support for other models doesn't make the driver name confusing.

PR #207, opened 2023-02-11
Add g_free after g_array_free if gboolean is False

Fr3ya:
In Glib's [g_array_free](https://docs.gtk.org/glib/type_func.Array.free.html) document, it said" _The element data if free_segment is FALSE, otherwise NULL. The element data should be freed using g_free()._ " So if the second parameter of g_array_free is FALSE, the array's element, i.e., array->data should be freed by g_free.

### DIFF
--- a/src/drivers.c
+++ b/src/drivers.c
@@ -54,4 +54,5 @@ SR_API void sr_drivers_init(struct sr_context *ctx)
 #endif
 	ctx->driver_list = (struct sr_dev_driver **)array->data;
 	g_array_free(array, FALSE);
+	g_free(array->data);
 }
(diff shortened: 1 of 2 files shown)


fenugrec:
I think this is interesting... GIMP has occurences of both :

https://gitlab.gnome.org/search?search=g_array_free&nav_source=navbar&project_id=1848&group_id=8&search_code=true&repository_ref=master

and when called with free_segment=FALSE, the return value of `g_array_free` is kept for whatever use it has.

The IBM tutorial linked above doesn't explicitly discuss free-ing the array, seems to be intended more as  a general-purpose overview, and their usage is sortof correct since on exiting main() everything is free`d anway.

fenugrec:
For future reference, in the first instance (in drivers.c) no, we don't to free the data because on the line just above

`ctx->driver_list = (struct sr_dev_driver **)array->data;`

we need that data for the driver list;

WIP

fenugrec:
Merged as of c78fa9cf9416260e018c131f72e9115302deba7f

PR #208, opened 2023-02-12
Add Atorch driver

pilatomic:
This adds the driver for some Atorch devices, namely :
- J7-c USB voltage / current / power / energy meter
- DL24MP digital load

It modifies the function provided by "binary_helpers.c" in 2 ways : 
1. Add support for values of 24 bits length
2. Add support for sr_mqflags ( seems needed for SmuView to exploit the signals )


fenugrec:
I haven't looked at all the changeset but one thing caught my attention. The commit "Added 24 bits binary helpers" is problematic because it bundles unrelated changes in it:

- you add the read24be implementation as advertised, this is ok
- replaced `sizeof(uintXX_t))` with the expected length; not sure why you're doing that at this time. To be fair, I'm not sure why we're using `sizeof(....` in the first place . I probably also would have written it as explicit lengths instead of sizeof. Still, IMO that doesn't belong in a commit with this message.
- the main issue : you replaced `sizeof(float)` with `4` . As you know `float` is not guaranteed to be 4 bytes on all platforms. Was this causing problems in existing code ? If so, it's an entirely different discussion that has nothing to do with the Atorch driver. I don't know the implications of changing it but would be very cautious of breaking other drivers (or certain platforms), and would at the very least expect a separate commit with a very good explanation of the rationale.

Note : I am neither a dev nor maintainer of sigrok, these are my own opinions as a spectator.

pilatomic:
Thank you for your input !
I talked a bit on IRC about this PR, and I will rework the whole patchset.
I indeed believe that for use of sizeof() is misguided use, as those value represent the size on the wire, and not the size on the host machine. That is why replacing float by 4 makes sense, the value is always 4  byte long on the wire.

pilatomic:
Updated to reflect all the issues pointed by reviewers : 
- Changed author name
- Commit messages reworked to match project approach
- Fixed style issues
- Use helper for conn
- Does not break compilation anymore when serial not supported
- Drop comments / dead code
- Reworked parser to operate on whole buffer and not one byte at at time ( previous version read / parsed byte by byte )
- The changes to binary_helpers.c are now slit into different commit, so they can be discussed individually

PR #210, opened 2023-02-14
Add long command processing time for Velleman PS3005DV1.3

pilatomic:
The Velleman PS3005D V1.3 power supply ( driver korad_kaxxxxp ) seems to
have a quirky RS232 interface, that requires waiting a longer time after
some commands before being allowed to send a new command.
Failing to respect this delay means the new command will NOT be executed.
This is especially problematic as the actual driver would report "0.000" value
for current when the power supply failed to answer the IOUT? request,
despite the output current being non zero.

This commit adds a quirk for this specific revision, and introduces a longer
delay for the affected commands ( STATUS? + any set command ).
The behavior is unchanged for other models.

### DIFF
--- a/src/hardware/korad-kaxxxxp/protocol.c
+++ b/src/hardware/korad-kaxxxxp/protocol.c
@@ -21,7 +21,11 @@
 #include <config.h>
 #include "protocol.h"
 
-#define DEVICE_PROCESSING_TIME_MS 80
+#define DEVICE_STD_PROCESSING_TIME_MS	80
+#define DEVICE_LONG_PROCESSING_TIME_MS	500
+
+#define STD_PROCESSING_TIME		0
+#define LONG_PROCESSING_TIME		1
 
 SR_PRIV int korad_kaxxxxp_send_cmd(struct sr_serial_dev_inst *serial,
 				const char *cmd)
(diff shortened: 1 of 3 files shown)


src/hardware/korad-kaxxxxp/protocol.c
```diff
@@ -155,15 +159,28 @@ static void give_device_time_to_process(struct dev_context *devc)
...
 		sr_spew("Sleeping for processing %" PRIi64 " usec", sleeping_time);
 	}
 }
 
+static void store_next_req_time(struct dev_context *devc, int request_long_processing_time)
+{
+	int64_t processing_time_us;
+
+	/* Some devices requires a longer "cooldown" time after some commands */
+	if(devc->model->quirks & KORAD_QUIRK_LONG_PROCESSING_TIME &&
```
knarfS, line 175: Style nit: Space between `if` and `(`

src/hardware/korad-kaxxxxp/protocol.c
```diff
@@ -305,7 +322,10 @@ SR_PRIV int korad_kaxxxxp_get_value(struct sr_serial_dev_inst *serial,
 		return ret;
 	}
 
-	devc->req_sent_at = g_get_monotonic_time();
+	if(target == KAXXXXP_STATUS)
```
knarfS, line 325: Style nit: space between `if` and `(`

knarfS (changes requested):
Thanks for your PR.
Apart from two style nits, everything looks good to me.

pilatomic:
Thank you for your review, I have fixed the 2 lines you pointed out.

PR #211, opened 2023-03-09
Add support for Fluke 87/89 DMM

pilatomic:
Those uses the same protocol as Fluke 187 / 189, so simply add their model number.
Also, do not consider 0.0 as an invalid measurement anymore, as it can be a perfectly valid measurement.
Tested working with a Fluke 87 DMM


pilatomic:
Update this PR to extract digit count from DMM response instead of using arbitrary value. Also propagate the 0.0 measurement invalid fix to 28x parsing function.

pilatomic:
Merged

PR #213, opened 2023-03-11
Driver for select Tektronix TDS, TPS, and TBS low-end models

byteit101:
I added the `tektronix-ocp2k5` driver for TBS1000B/EDU, TBS1000, TDS2000C/TDS1000C-EDU, TDS2000B/TDS1000B, TDS2000/TDS1000, TDS200 and TPS2000B/TPS2000 Series Digital Oscilloscopes. I tried to find a proper name for this protocol, but Tektronix support was being squirrely about a name, so I just arbitrarily came up with the blandly-generic "oscilloscope control protocol 2k5" as that is the only words somewhat used to describe it in the programming manual, plus the size of memory of all of these devices. I also thought about calling it the "lem2k5" for "Low-end Millennium" as this protocol seems to emerge around the turn of the millennium in low-end scopes with the TDS200 series. I'm happy to oblige a name change to something else.

I have verified it using a [TDS2000B series scope](https://sigrok.org/wiki/Tektronix_TDS2000B_series). A lot of features aren't supported, but I think the critical ones are taken care of. Luckily, since it appears that local control isn't locked out, more advanced features can still be configured via the front panel. The most glaring example of this is the trigger voltage, which PulseView doesn't seem to support.

I didn't update the udev rules, but I am querying the vendor for all the relevant PIDs. This branch is based on the sigrok.org master, not the out-of-date github master.

Here are three relevant screenshots of pulseview with this driver:

tek_settings

tek_channel_settings

tek_channels

Finally, the timescale is 1-2.5-5, so the rational is not always power of 1000, leading to this slight weirdness in PulseView, but I think it should be fine?

tek_timescapes


byteit101:
> I didn't update the udev rules, but I am querying the vendor for all the relevant PIDs. 

The vendor got back to me and was unable to provide such a list

pilatomic:
Hello @byteit101 ,
I haven't looked at your code yet, but a quick remark is that you should update the commit messages to follow the project template (prefix each message with the name of the driver touched), and maybe squash some of the cleanup / smaller commits.

byteit101:
Ah, I had assumed it would be when it was merged, but in case, not, I squashed everything down.

If you are looking for the previous history, it kept that on my tek_ocp2k5 branch

pilatomic:
Thats a bit too much squash x)

Good practice here it to have at least a first "Initial driver skeleton" commit, and then only the actual code. You can of course have more commits, just ensure that hey are individually build-able to allow bisection.

Also check that your code matches the project code styling.

Cannot be to much more specific, I'm also a new guy around here.

byteit101:
Ah, well, I added it in a different folder and renamed it late in the game, and attempting to re-create a fake history sounds tedious. 

I did format it (except the table) with @knarfS 's clang-format patch/branch, let me know if something isn't quite right

byteit101:
I have pulled it slightly apart again, though also renamed it to `tektronix-tds` as `ocp2k5` was a bit too clunky

byteit101:
After review from `gsi` I've updated the driver to use a mutex, fixed the digit values, and reordered protocol.c

v0lker:
sorry for butting in, but i'm interested in using your branch with my TDS2022 (via USB/RS-232 converter) and while it seems to talk to it to some extent, there seems to be an issue:
```
 $ sudo /usr/local/bin/sigrok-cli -l 5 --driver=tektronix-tds:conn=/dev/ttyUSB0 --show
sr: [00:00.000017] log: libsigrok loglevel set to 5.
sr: [00:00.000042] backend: libsigrok 0.6.0-git-4ea2dce7/4:0:0.
sr: [00:00.000085] backend: Libs: glib 2.72.4 (rt: 2.72.4/7204:4), zlib 1.2.11, libzip 1.7.3, minilzo 2.10, libserialport 0.1.1/1:0:1 (rt: 0.1.1/1:0:1), libusb-1.0 1.0.25.11696 API 0x01000109.
sr: [00:00.000096] backend: Host: x86_64-pc-linux-gnu, little-endian.
sr: [00:00.000103] backend: SCPI backends: TCP, RPC, serial, USBTMC.
[...]
sr: [00:00.007714] hwdriver: sr_config_list(): key 2147418112 (NULL) sdi (nil) cg NULL -> [uint32 20000, 20001]
Scan options:
    conn
    serialcomm
sr: [00:00.008012] hwdriver: sr_config_list(): key 2147418112 (NULL) sdi (nil) cg NULL -> [uint32 20000, 20001]
sr: [00:00.008031] serial: Parsed serial device: /dev/ttyUSB0.
sr: [00:00.008044] scpi: Opening serial device /dev/ttyUSB0.
sr: [00:00.013652] serial: Opening serial port '/dev/ttyUSB0' (flags 1).
sr: [00:00.015141] serial: Flushing serial port /dev/ttyUSB0.
sr: [00:00.015291] serial: Wrote 6/6 bytes.
sr: [00:00.015298] scpi_serial: Successfully sent SCPI command: '*IDN?'.
sr: [00:00.067738] serial: Read 1/2048 bytes.
[...]
sr: [00:00.099545] serial: Read 1/1991 bytes.
sr: [00:00.099552] scpi_serial: Received NL terminator
sr: [00:00.099584] scpi: Got response: 'TEKTRONIX,TDS 2022,0,CF:91.1CT FV:v2.60 TDS2CM:CMV:v1.04', length 56.
sr: [00:00.099676] serial: Closing serial port /dev/ttyUSB0.
sr: [00:00.239022] hwdriver: Scan found 1 devices (tektronix-tds).
tektronix-tds - Tektronix TDS 2022 CF:91.1CT FV:v2.60 TDS2CM:CMV:v1.04 [S/N: 0] with 2 channels: CH1 CH2
sr: [00:00.239106] device: tektronix-tds: Opening device instance.
sr: [00:00.239151] serial: Opening serial port '/dev/ttyUSB0' (flags 1).
sr: [00:00.241578] serial: Flushing serial port /dev/ttyUSB0.
sr: [00:00.241706] serial: Wrote 12/12 bytes.
sr: [00:00.241720] scpi_serial: Successfully sent SCPI command: 'SELECT:CH1?'.
sr: [00:30.241735] scpi: Timed out waiting for SCPI response.
sr: [00:30.241750] tektronix-tds: Failed to get device config: generic/unspecified error.
Failed to open device
```

interestingly, the scope seems to reply when i try to reproduce that with picocom:
```
$ picocom --echo -b 19200 /dev/ttyUSB0
picocom v3.1

port is        : /dev/ttyUSB0
flowcontrol    : none
baudrate is    : 19200
parity is      : none
databits are   : 8
stopbits are   : 1
escape is      : C-a
local echo is  : yes
noinit is      : no
noreset is     : no
hangup is      : no
nolock is      : no
send_cmd is    : sz -vv
receive_cmd is : rz -vv -E
imap is        : 
omap is        : 
emap is        : crcrlf,delbs,
logfile is     : none
initstring     : none
exit_after is  : not set
exit is        : no

Type [C-a] [C-h] to see available commands
Terminal ready
SELECT:CH1?
1
SELECT:CH2?
0

Terminating...
Thanks for using picocom
```


i tried increasing the time-outs, adding line delimiters on the libsigrok end and changing them on the scope (i think `\n` only is the correct setting?)   --  all  to no avail so far.

any suggestions? thanks!

byteit101:
@v0lker That's interesting. I know `gsi` ran into similar timeout issues on a very old model, and his fixes caused timeout issues for me. I didn't understand sigrok timeouts well enough to solve the issue, and I've been extremely busy for the past year, so I unfortunately don't have much time to dig further. I can definitely help test still, though it may take some time.

Have you tried wireshark/usb/serial captures of failures vs picocom?

biot:
This is not a timeout issue: the log clearly shows it waiting for 30 seconds before timing out.

I think it's a line termination issue. libsigrok's SCPI layer just adds `\n` to commands sent out if not already there, and that's not working for @v0lker. But the picocom session emap set to `crcrlf`, meaning translate `cr` to `crlf`, i.e. `\r\n`.

If you add `\r` to your command strings in `protocol.c`, i think that will make it work. However it's clear that @byteit101's TDS model doesn't need the extra `\r`, but will it work with that added?

v0lker:
> If you add `\r` to your command strings in `protocol.c`, i think that will make it work. 

as stated above, i tried that already. just went through the logs that i captured and unfortunately i didn't keep those experiments.
as

byteit101:
> i think a big difference between how @byteit101 and i are connecting is that he's using USBTMC on the TDS2000B, whereas i'm using a TDS2022 + comms module with an RS-232 interface (and connecting to it via UBS/RS-232 converter), thus it's connecting via serial.

This is likely the difference, IIRC `gsi` was using a `TDS200` model, also with a RS-232 interface. I only had the USBTMC model, so didn't test RS-232 at all

PR #215, opened 2023-03-30
fluke-dmm: Fix significant digits count

pilatomic:
This fixes the fluke-dmm having an arbitrary, hard-coded significant digits count.

The first commit  fixes an oversight in a previous patch, where the measurement exponent was not taken into consideration to get the actual significant digits count.

The other commits apply the same logic to the 28x code path.
I do not have a Fluke 28x DMM to test, although I did test it against the examples provided in the "Fluke 289/287 Remote Interface Specification" (using an external USB device programmed to answer to the "ID" and "QM" commands), and it performed as expected : 

Response to QM command : "0\r-0.023E-3,VDC,NORMAL,NONE\r"
sigrok-cli displays : P1: -23 µV DC

"0\r0.255E-3,VAC,NORMAL,NONE\r" : P1: 255 µV AC RMS
"0\r9.323E0,VDC,NORMAL,NONE\r" : P1: 9.323 V DC
"0\r+9.99999999E+37,VDC,OL,NONE\r" : P1: nan V
"0\r58.99E0,VAC,NORMAL,NONE\r" : P1: 58.99 V AC RMS
"0\r63.679E0,Hz,NORMAL,POSITIVE_EDGE\r" : P1: 63.679 Hz
"0\r262.39E-3,VAC,NORMAL,NONE\r" : P1: 262.39 mV AC RMS
"0\r75.0E0,FAR,NORMAL,NONE\r" : P1: 75.0 °F
"0\r23.9E0,CEL,NORMAL,NONE\r" : P1: 23.9 °C
"0\r50.75E0,OHM,NORMAL,NONE\r" : P1: 50.75 Ω
"0\r50.762E0,OHM,NORMAL,NONE\r" : P1: 50.762 Ω
"0\r+9.99999999E+37,OHM,OL,NONE\r" : P1: inf TΩ
"0\r0.95E-6,F,NORMAL,NONE\r" : P1: 0.95 µF
"0\r0.5498E0,VDC,NORMAL,GOOD_DIODE\r" : P1: 549.8 mV DC
"0\r979.0E-6,ADC,NORMAL,NONE\r" : P1: 979.0 µA DC
"0\r1.000E-3,ADC,NORMAL,NONE\r"P1: 1000 µA DC

Only this entry is not processed by libsigrok, but neither is it without this PR. This will require a subsequent patch.
"0\r0.2785E0,VAC_PLUS_DC,NORMAL,NONE\r"

This should provide enough confidence to merge this PR.

I would still like to have some real world testing of that though, so if anyone with a Fluke 28x could chime in, that would be great !


pilatomic:
Merged into master

PR #217, opened 2023-04-11
Atorch rev4

pilatomic:
Revision 4 of my Atorch driver

Tested with with following devices :

-  J7-c USB voltage / current / power / energy meter

- DL24MP digital load


src/binary_helpers.c
```diff
@@ -62,3 +62,54 @@ SR_PRIV int bv_get_value_with_length_check(float *out, const struct binary_value
 		*out = value;
 	return SR_OK;
 }
+
+SR_PRIV int bv_get_value(float *out, const struct binary_value_spec *spec,
+	const void *data)
```
dresco, line 67: Should the void* here also be changed to a uint8_t* as it was for the _with_length_check version?
Currently generates a compiler warning on dereferencing the void pointer..

src/binary_helpers.c
```diff
@@ -41,11 +41,13 @@ SR_PRIV int bv_get_value(float *out, const struct binary_value_spec *spec,
 	VALUE_TYPE(BVT_UINT8, read_u8, sizeof(uint8_t));
 
 	VALUE_TYPE(BVT_BE_UINT16, read_u16be, sizeof(uint16_t));
+	VALUE_TYPE(BVT_BE_UINT24, read_u24be, 3);
```
dresco, line 44: Honestly not sure, but would it be more in keeping to make these 3 * sizeof(uint8_t) ?

pilatomic:
Thank you for pointing this out, this has been fixed in the rev5

PR #108, opened 2020-12-03
Add driver for BK Precision 1856D

danselmi:

PR #218, opened 2023-04-24
Add driver for BK Precision 1856D

danselmi:
continues https://github.com/sigrokproject/libsigrok/pull/108 which was on master. Changed to separate branch here.


src/hardware/bk-1856d/protocol.c
```diff
@@ -0,0 +1,375 @@
...
+{
+	int zero_found;
+	int has_data;
+
+	zero_found = 0;
+	has_data = 0;
+
+	for (int idx = 0; idx < BK1856D_MSG_SIZE - 1; ++idx)
+	{
+		if (devc->buffer[idx] == ' ')
```
fenugrec, line 197: this chain of if / else if all need braces

src/hardware/bk-1856d/protocol.c
```diff
@@ -0,0 +1,375 @@
...
+
+	g_mutex_lock(&devc->rw_mutex);
+	bk_1856d_send_input_sel(sdi);
+
+	bk_1856d_send_gate_time(sdi);
+	bk_1856d_request_data(sdi);
+	g_mutex_unlock(&devc->rw_mutex);
+}
+
+static int bk_1856d_check_for_zero_message(struct dev_context *devc)
```
fenugrec, line 187: it would be helpful to the reviewer (present and future) to give a few examples, in a comment, of typical strings and expected return values.

src/hardware/bk-1856d/protocol.c
```diff
@@ -0,0 +1,375 @@
...
+static void bk_1856d_parse_message(const struct sr_dev_inst *sdi)
+{
+	struct dev_context *devc;
+	struct sr_serial_dev_inst *serial;
+
+	double freq_value;
+	int digits;
+	char *endPtr, *dotPtr;
+
+	if (!(devc = sdi->priv) || !(serial = sdi->conn))
```
fenugrec, line 246: assignment in conditional

src/hardware/bk-1856d/protocol.c
```diff
@@ -0,0 +1,375 @@
...
+		return;
+
+	/* check for cr at end of message */
+	if (devc->buffer[BK1856D_MSG_SIZE-1] != '\xD')
```
fenugrec, line 250: remember '\r' already gives you a carriage-return (0x0D);  IMO slightly easier to read ? This applies to the list of command defines earlier too.

src/hardware/bk-1856d/protocol.c
```diff
@@ -0,0 +1,375 @@
...
+		bk_1856d_request_data(sdi);
+		g_mutex_unlock(&devc->rw_mutex);
+		return;
+	}
+
+	devc->buffer[BK1856D_MSG_SIZE-1] = 0; /* set trailing zero */
+
+	if (bk_1856d_check_for_zero_message(devc))
+	{
+		sr_spew("received 0 package");
```
fenugrec, line 267: suggest  rewording, otherwise can be misinterpreted as "we received no packages". Maybe "empty packet", or "zero message", or something else ?

src/hardware/bk-1856d/protocol.c
```diff
@@ -0,0 +1,375 @@
...
+		devc->buffer_level = 0;
+		bk_1856d_request_data(sdi);
+		return;
+	}
+
+	sr_dbg("received msg: '%s'", devc->buffer);
+	freq_value = strtod(devc->buffer, &endPtr);
+	sr_dbg("parsed value: %f", freq_value);
+
+	if (strcmp(devc->buffer + BK1856D_MSG_NUMBER_SIZE+1, "Hz ") != 0)
```
fenugrec, line 277: this looks like it needs bounds checking on devc->buffer, I haven't seen anything that guarantees a length >= (BK1856D_MSG_NUMBER_SIZE + 1 + strlen("Hz "))

src/hardware/bk-1856d/protocol.c
```diff
@@ -0,0 +1,375 @@
...
+	{
+		devc->buffer_level = 0;
+		g_mutex_lock(&devc->rw_mutex);
+		bk_1856d_chk_select_input(sdi);
+		bk_1856d_send_gate_time(sdi);
+		bk_1856d_request_data(sdi);
+		g_mutex_unlock(&devc->rw_mutex);
+	}
+	else
+		sr_dev_acquisition_stop((struct sr_dev_inst *)sdi);
```
fenugrec, line 318: you are explicitly casting to bypass the 'const' attribute, this never feels right. I see other drivers also do this, would like to hear input from others about this.

src/hardware/bk-1856d/protocol.c
```diff
@@ -0,0 +1,375 @@
...
+SR_PRIV int bk_1856d_receive_data(int fd, int revents, void *cb_data)
+{
+	const struct sr_dev_inst *sdi;
+	struct dev_context *devc;
+	struct sr_serial_dev_inst *serial;
+	int len;
+
+	(void)fd;
+
+	if (!(sdi = cb_data) || !(devc = sdi->priv))
```
fenugrec, line 330: assignment in conditional

src/hardware/bk-1856d/protocol.c
```diff
@@ -0,0 +1,375 @@
...
+		sr_dbg("received timeout");
+		g_mutex_lock(&devc->rw_mutex);
+		bk_1856d_send_gate_time(sdi);
+		bk_1856d_request_data(sdi);
+		g_mutex_unlock(&devc->rw_mutex);
+
+		return TRUE;
+	}
+
+	if (!(serial = sdi->conn))
```
fenugrec, line 348: assignment in conditional

fenugrec (commented):
I'm not too sure about the usage of the mutex here. Unclear which functions can be called from different threads, and how the changes (e.g. to `gate_time` or `sel_input` are propagated asynchronously.

src/hardware/bk-1856d/protocol.c
danselmi: I know, the idea to use \xD is because in the vendors manual 0DH is used to indicate the "Terminate Code".
If you preferre, I will change it to \r. Let me know.

src/hardware/bk-1856d/api.c
```diff
@@ -0,0 +1,247 @@
...
+
+	(void)cg;
+
+	if (!(devc = sdi->priv))
+		return SR_ERR;
+
+	switch (key) {
+	case SR_CONF_TIMEBASE:
+		{
+			uint64_t p, q;
```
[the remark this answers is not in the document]
danselmi: On frequency meters one usually configures the gate time  during which the number of periods is counted).
I believe SR_CONF_TIMEBASE is the parameter for this - not?

This has nothing to do with the protocol but maps the 4 supported gate times to the coresponding commands.

src/hardware/bk-1856d/api.c
```diff
@@ -0,0 +1,247 @@
...
+	sdi->inst_type = SR_INST_SERIAL;
+	sdi->priv = devc;
+	g_mutex_init(&devc->rw_mutex);
+	sr_channel_new(sdi, 0, SR_CHANNEL_ANALOG, TRUE, "P1");
+	devc->sel_input = InputC;
+	devc->curr_sel_input = InputC;
+	devc->gate_time = 0;
+	devc->hold = Off;
+
+	g_usleep(150 * 1000); /* Wait a little to allow serial port to settle. */
```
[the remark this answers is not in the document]
danselmi: I found no other solution than this (same for the sleep time after changing the gate time).
If I remember correctly (it's more than two years since), I found the values empirically and rounded up.

src/hardware/bk-1856d/protocol.c
danselmi: I added a link to the vendors manual.

src/hardware/bk-1856d/protocol.c
```diff
@@ -0,0 +1,375 @@
...
+			break;
+		case 1:
+			cmd = GATE_TIME_1;
+			sr_info("sending gate time 1 (100ms)");
+			microseconds = 80000;
+			break;
+		case 2:
+			cmd = GATE_TIME_2;
+			sr_info("sending gate time 2 (1s)");
+			microseconds = 80000;
```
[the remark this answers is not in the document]
danselmi: About the time: see the comment for the sleep at the end of api.c::scan(...).
 
Changed to array of structs. 
They are only used in this function... I am not sure if this is the "nicer" solution ;)

danselmi:
> I'm not too sure about the usage of the mutex here. Unclear which functions can be called from different threads, and how the changes (e.g. to `gate_time` or `sel_input` are propagated asynchronously.

Me neither. It's implementd after a discussion with knarfS. Probably smuview calls them from different threads.

src/hardware/bk-1856d/protocol.c
danselmi: Actually bk_1856d_parse_message() is only called when the buffer is full.

src/hardware/bk-1856d/api.c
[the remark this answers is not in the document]
biot: But you haven't actually sent anything to the device yet, `devc->gate_time` is just a bit of memory. Why would you need a delay?

biot:
- I think you should add a new config key, SR_CONF_GATE_TIME. This is not a time base, and in fact a time base has a very different meaning in frequency counters.
- i agree with you and fenugrec that those mutexes are wrong. They're not only in the wrong place in this driver, I don't think mutexes belong in drivers at all. If needed, they belong in the application, which has full control over when in calls into libsigrok. Need to have a conversation with @knarfS about it.

src/hardware/bk-1856d/protocol.c
```diff
@@ -17,25 +17,344 @@
...
+	if (!sdi)
+		return;
+
+	devc = sdi->priv;
+	serial = sdi->conn;
+	if (!devc || !serial)
+		return;
+
+	if (devc->gate_time > 3)
+		devc->gate_time = 3;
```
biot, line 126: This seems unnecessary, since the code that sets this can never exceed bounds.

src/hardware/bk-1856d/protocol.c
```diff
@@ -17,25 +17,344 @@
...
+
+static int bk_1856d_check_for_zero_message(struct dev_context *devc)
+{
+	int zero_found;
+	int has_data;
+
+	zero_found = 0;
+	has_data = 0;
+
+	for (int idx = 0; idx < BK1856D_MSG_SIZE - 1; ++idx) {
```
biot, line 190: Please put all variable declarations up top.

src/hardware/bk-1856d/protocol.c
```diff
@@ -17,25 +17,344 @@
...
+	devc = sdi->priv;
+	serial = sdi->conn;
+
+	/* check for cr at end of message */
+	if (devc->buffer[BK1856D_MSG_SIZE-1] != '\xD') {
+		sr_err("expected cr at end of message.");
+		devc->buffer_level = 0;
+		serial_flush(serial);
+		bk_1856d_send_input_sel(sdi);
+		bk_1856d_send_gate_time(sdi);
```
biot, line 251: You seem to select input and send gate time every time you interact with the device. Is that actually necessary? Do things break when you leave these out?

src/hardware/bk-1856d/protocol.c
```diff
@@ -17,25 +17,344 @@
...
+	if (bk_1856d_check_for_zero_message(devc)) {
+		sr_spew("received an empty packet");
+		devc->buffer_level = 0;
+		bk_1856d_request_data(sdi);
+		return;
+	}
+
+	sr_dbg("received msg: '%s'", devc->buffer);
+	freq_value = strtod(devc->buffer, &endPtr);
+	sr_dbg("parsed value: %f", freq_value);
```
biot, line 267: Once your parser works, you can drop these low-level debug statements.

src/hardware/bk-1856d/protocol.c
biot: This shouldn't be cast because it shouldn't be `const` in the first place. The mistake is in the next function.

src/hardware/bk-1856d/protocol.c
```diff
@@ -17,25 +17,344 @@
...
+		bk_1856d_send_gate_time(sdi);
+		bk_1856d_request_data(sdi);
+	}
+	else
+		sr_dev_acquisition_stop((struct sr_dev_inst *)sdi);
+}
+
 SR_PRIV int bk_1856d_receive_data(int fd, int revents, void *cb_data)
 {
 	const struct sr_dev_inst *sdi;
```
biot, line 309: There it is. This shouldn't be `const`.

src/hardware/bk-1856d/protocol.c
```diff
@@ -17,25 +17,344 @@
...
-	if (!(devc = sdi->priv))
+	if (!sdi->priv)
 		return TRUE;
+	devc = sdi->priv;
+
+	if (revents != G_IO_IN) {
+		/* Timeout
+		   in rare cases th bk1856 doesn't respond anymore
+		   (probably timing on rs232)
+		   for now, just restart the measurement. */
```
biot, line 328: Please reformat this into three lines.

biot (changes requested):
Mostly coding style fixes needed, driver looks good otherwise.

biot:
I forgot to add, can you also rename the driver? These generally follow the pattern `vendor-model` where possible, so ` bkprecision-1856d` would be better. Thanks!

biot (approved):
Thanks for the updates, looks great!

abraxa:
Merged as https://sigrok.org/gitaction/libsigrok.git/1b12204..73d1e16, thanks for your continued contributions!

PR #219, opened 2023-05-30
Support for Korad KKG305P

IvanBayan:
Support for KKG305P with new quirk.
korad_kaxxxxp_send_cmd() changed in a way to support new line termination.
STATUS parsed according to Korad "documentation".
New flags in device structure to support status of remote sensing status and external control interface status.


abraxa:
Hi, thanks for your contribution! Would you be so kind and rebase your branch on top of current git head to solve the conflicts?

IvanBayan:
I will try and let you know.

src/hardware/korad-kaxxxxp/protocol.h
```diff
@@ -79,16 +80,20 @@ struct dev_context {
...
+	gboolean ovp_enabled;      /**< Output voltage protection enabled. */
+	gboolean rmt_comp_enabled; /**< Is remote compensation enabled. */
+	gboolean ext_int_enabled;  /**< Is external interface enabled. */
 
 	gboolean cc_mode_1_changed;      /**< CC mode of channel 1 has changed. */
 	gboolean cc_mode_2_changed;      /**< CC mode of channel 2 has changed. */
 	gboolean output_enabled_changed; /**< Output enabled state has changed. */
 	gboolean ocp_enabled_changed;    /**< OCP enabled state has changed. */
 	gboolean ovp_enabled_changed;    /**< OVP enabled state has changed. */
+	gboolean rmt_comp_changed;       /**< Is remote compensation enabled. */
```
fenugrec, line 95: wrong copypasted comment here and below, `enabled`-> `changed`

fenugrec (commented):
(note I am neither a main dev or maintainer of sigrok. Just a 'concerned citizen')

I don't see major problems, but I wonder about the choice of changing every call to `korad_kaxxxxp_send_cmd()` with the extra arg.
I think I would've implemented a new `korad_kkg_send_cmd()` with the newline fixes (but no additional argument); change the `korad_kaxxxxp_send_cmd()` call sites to a instead call a *function pointer, and set that function pointer once at device init according to device type. Then after writing that I would then start doubting on whether it was a even a good idea.

Hope we get some input from others about this

fenugrec:
Also forgot to mention, one of these commits `[Fix un-conditional double read](https://github.com/sigrokproject/libsigrok/pull/219/commits/d4726c18019c5b84d1b7458e0a3bcfbebbf5f592) ` will need a better commit message

IvanBayan:
Rebased. Includes small fix for debug messages.

PR #220, opened 2023-06-20
uni-t-ut8802e: Initial commit

KonradTagowski:
Implementation of the driver for Uni-t UT8802E bench multimeter. The protocol was reconstructed manually. I suppose that the multimeter supports only reading the measurement.


src/hardware/uni-t-ut8802e/protocol.c
```diff
@@ -0,0 +1,430 @@
...
+    feedbuff.analog.meaning->mq = info->meas_head.mode;
+    feedbuff.analog.meaning->unit = info->meas_data.main_unit;
+
+    info->meas_data.main_value = new_value * pow(10, info->meas_data.main_prec);
+
+    feedbuff.analog.encoding->digits = -info->meas_data.main_prec + 4;
+    feedbuff.analog.spec->spec_digits = -info->meas_data.main_prec + 4;
+
+    feedbuff.main_value = info->meas_data.main_value;
+	if (sdi->status != SR_ST_ACTIVE)
```
harper23, line 299: I don't see any reason for this indentation.

src/hardware/uni-t-ut8802e/protocol.c
```diff
@@ -0,0 +1,430 @@
...
+ * - UART frame format 8n1 at 9600 bps.
+ * - DMM packet starts with a magic marker, functionality, value, etc.
+ * - UT8802E only sends a packet with measurement. Example of frame:
+ *   Frame | func | value    | comma | settings | checksum |
+ *   ac    | 1b   | 45 01 00 | 33	 | 04  	    | 44       |
+ * - The frame to send a command to the multimeter is unknown.
+*/
+
+static const int8_t range_volt[]  = {-3,  0};
+static const int8_t range_amp[]   = {-6, -3,  0};
```
harper23, line 37: range_amp[2] is never used. The initializer list can be shortened.

src/hardware/uni-t-ut8802e/protocol.c
```diff
@@ -0,0 +1,430 @@
...
+    clc_checksum = 0;
+    checksum_dlen = got_length;
+    for (uint32_t i = 0; i < checksum_dlen; i++) {
+        clc_checksum = clc_checksum + pkt[i];
+    }
+    want_checksum = clc_checksum;
+
+    if (clc_checksum > 256 && want_checksum >= 128)
+        want_checksum = want_checksum - 128;
+
```
harper23, line 110: The checksum is calculated over the first 8 bytes including the FRAME_MAGIC byte (0xAC).  If the checksum is in the range 0xAC - 0xFF, i.e. the sum of the remaining bytes is less than 0x54, the checksum should be  found in pkt[8] with bit 7 set. But if the checksum is 0x100 or more, bit 7 of the checksum will be reset.
This is a strange definition.

KonradTagowski: Thank you very much. I made a mistake here. I fixed it now. Bit 7 is always cleared.
0xac 0x04 0x10 0x20 0x01 0x33 0x04 0x18 -> Sum of bytes 0x118,
0xac 0x06 0x00 0x00 0x00 0x31 0x04 0x67 (103) -> Sum of bytes 0xE7 (231) 
0xac 0x04 0x88 0x99 0x01 0x33 0x04 0x09 -> Sum of bytes 0x209
0xac 0x04 0x89 0x69 0x01 0x33 0x04 0x5a (90)-> Sum of bytes 0x1DA (474)

src/hardware/uni-t-ut8802e/protocol.c
KonradTagowski: Range_amp[2], I used to choose a range for the current measurement in 168 to 194 lines.

src/hardware/uni-t-ut8802e/protocol.c
KonradTagowski: Fixed

src/hardware/uni-t-ut8802e/api.c
```diff
@@ -0,0 +1,219 @@
...
+    case SR_CONF_LIMIT_FRAMES:
+    case SR_CONF_LIMIT_SAMPLES:
+    case SR_CONF_LIMIT_MSEC:
+        if (!devc)
+            return SR_ERR_ARG;
+        return sr_sw_limits_config_get(&devc->limits, key, data);
+    case SR_CONF_DATA_SOURCE:
+        *data = g_variant_new_string(data_sources[0]);
+        break;
+    // TO DO ADD SR_CONF_MEASURED_QUANTITY and SR_CONF_RANGE
```
biot, line 125: Please don't leave TODOs in the code that's going up to mainline. It's ok if your driver doesn't yet support all features the hardware has, but the driver should just be functional for the stuff it does support. You can make a note of features it still needs in the commit message or (much preferred) the wiki.

src/hardware/uni-t-ut8802e/api.c
```diff
@@ -0,0 +1,219 @@
...
+static const uint32_t devopts[] = {
+    SR_CONF_CONTINUOUS,
+    SR_CONF_LIMIT_SAMPLES | SR_CONF_GET | SR_CONF_SET,
+    SR_CONF_LIMIT_MSEC | SR_CONF_GET | SR_CONF_SET,
+    SR_CONF_DATA_SOURCE | SR_CONF_GET | SR_CONF_LIST,
+};
+
+
+static const char *channel_names[] = {
+    "Main",
```
biot, line 40: Regular multimeters like this generally use `P1` as the channel name (for probe 1).

src/hardware/uni-t-ut8802e/api.c
```diff
@@ -0,0 +1,219 @@
...
+};
+
+static const char *data_sources[] = {
+    "Live",
+};
```
biot, line 45: Since this multimeter only has live measurements, i.e. no storage mode, there is no need to implement `SR_CONF_DATA_SOURCE` in this driver. So this can go, and the entry in `devopts[]`, and the `config_get()/config_list()` entries as well.

src/hardware/uni-t-ut8802e/api.c
```diff
@@ -0,0 +1,219 @@
...
+    sdi->priv = devc;
+    sr_sw_limits_init(&devc->limits);
+    for (idx = 0; idx < ARRAY_SIZE(channel_names); idx++) {
+        sr_channel_new(sdi, idx, SR_CHANNEL_ANALOG, TRUE,
+                channel_names[idx]);
+    }
+    devices = g_slist_append(devices, sdi);
+
+    return std_scan_complete(di, devices);
+
```
biot, line 99: Extra blank line.

src/hardware/uni-t-ut8802e/protocol.c
```diff
@@ -0,0 +1,430 @@
...
+/*
+ * Developer notes on the UT8802E protocol:
+ * - Serial communication over HID CP2110 converter USB-UART.
+ * - UART frame format 8n1 at 9600 bps.
+ * - DMM packet starts with a magic marker, functionality, value, etc.
+ * - UT8802E only sends a packet with measurement. Example of frame:
+ *   Frame | func | value    | comma | settings | checksum |
+ *   ac    | 1b   | 45 01 00 | 33	 | 04  	    | 44       |
+ * - The frame to send a command to the multimeter is unknown.
+*/
```
biot: This really belongs in the wiki, not in the code. Can you please create a page for this hardware? @abraxa will hook you up with an account. Document as much detail of the protocol as you can.

src/hardware/uni-t-ut8802e/protocol.c
biot: No, `range_amp[2]` is not used.

src/hardware/uni-t-ut8802e/protocol.c
```diff
@@ -0,0 +1,430 @@
...
+    float new_value;
+
+    struct feed_buffer feedbuff;
+
+    int ret;
+
+    devc = sdi ? sdi->priv : NULL;
+    info = devc ? &devc->info : NULL;
+
+    if (len < 1 * sizeof(uint8_t)) {
```
biot, line 82: This check seems wrong.

src/hardware/uni-t-ut8802e/protocol.c
```diff
@@ -0,0 +1,430 @@
...
+    got_magic_frame = R8(&pkt[0]);
+
+    if (got_magic_frame != FRAME_MAGIC) {
+        sr_spew("Wrong frame packet");
+        return SR_ERR_DATA;
+    }
+
+    got_length = 7;
+    if (got_length != len - sizeof(uint8_t)) {
+        sr_spew("Wrong lenght of packet");
```
biot, line 96: Misspelled length.

This whole packet validity check needs rewriting.

src/hardware/uni-t-ut8802e/protocol.c
```diff
@@ -0,0 +1,430 @@
...
+        return SR_ERR_DATA;
+    }
+
+    payload = &pkt[sizeof(uint8_t)];
+
+    clc_checksum = 0;
+    checksum_dlen = got_length;
+    for (uint32_t i = 0; i < checksum_dlen; i++) {
```
biot, line 104: Please declare variables at the top of the function.

src/hardware/uni-t-ut8802e/protocol.c
```diff
@@ -0,0 +1,430 @@
...
+        v8 = R8(&pkt[0]);
+        if (v8 != FRAME_MAGIC)
+            break;
+
+        pkt_len = 8;
+        sr_spew("Got a expected packet lenght %zu, have %zu",
+                pkt_len, devc->packet_len);
+
+        if (pkt_len > devc->packet_len)
+            return SR_OK;
```
biot, line 340: Surely there's no need to check packet validity both here and in `process_packet()`?

src/hardware/uni-t-ut8802e/protocol.c
```diff
@@ -0,0 +1,430 @@
...
+    struct dev_context *devc;
+    struct sr_serial_dev_inst *serial;
+    size_t len;
+    uint8_t *data;
+
+    devc = sdi->priv;
+    serial = sdi->conn;
+
+    if (devc->packet_len == sizeof(devc->packet)) {
+        (void)process_packet(sdi, &devc->packet[0], devc->packet_len);
```
biot, line 391: Why not use the return code? Might be an error.

src/hardware/uni-t-ut8802e/protocol.c
```diff
@@ -0,0 +1,430 @@
...
+
+    sdi = cb_data;
+    if (!sdi)
+        return TRUE;
+    serial = sdi->conn;
+    if (!serial)
+        return TRUE;
+
+    if (revents & G_IO_IN)
+        (void)ut8802e_receive_data(sdi);
```
biot, line 422: If you don't want to know if that function returned an error, just rewrite the function to not return one.

src/hardware/uni-t-ut8802e/protocol.h
```diff
@@ -0,0 +1,131 @@
...
+#include <stdint.h>
+#include <glib.h>
+#include <libsigrok/libsigrok.h>
+#include "libsigrok-internal.h"
+
+#define LOG_PREFIX "uni-t-ut8802e"
+
+#define PACKET_SIZE    32
+
+#define FRAME_MAGIC 0xac /* AC */
```
biot, line 32: The comment seems unnecessary.

src/hardware/uni-t-ut8802e/protocol.h
```diff
@@ -0,0 +1,131 @@
...
+
+enum ut8802e_rsp_type {
+    RSP_TYPE_INFO = 0x00,
+    RSP_TYPE_MEASUREMENT = 0x02,
+    RSP_TYPE_REC_INFO = 0x04,
+};
+
+enum ut8802e_channel_idx {
+    ut8802e_CH_MAIN,
+};
```
biot, line 91: There is only one channel, this seems unnecessary.

src/hardware/uni-t-ut8802e/protocol.c
```diff
@@ -0,0 +1,430 @@
...
+    for (uint32_t i = 0; i < checksum_dlen; i++) {
+        clc_checksum = clc_checksum + pkt[i];
+    }
+    want_checksum = clc_checksum;
+
+    if (clc_checksum > 256 && want_checksum >= 128)
+        want_checksum = want_checksum - 128;
+
+    got_checksum = R8(&pkt[checksum_dlen]);
+    sr_spew("Checksum: %d, Got: %d, ALL %d", want_checksum, got_checksum, clc_checksum);
```
biot: Once you know you've got the checksum right, no need to spew this debug output.

biot (changes requested):
Sorry for the late review, but here it is.

A bunch of mostly small comments, easily fixable. A couple of general notes:
- Your code uses spaces for indentation, please use tabs instead.
- Have you checked whether this protocol might already be supported by some other driver? This may well be driven by a generic DMM chip, and we already support lots of those (see https://sigrok.org/wiki/Multimeter_ICs). The easiest way to check is just to open up the device and check markings on the main IC.

PR #221, opened 2023-06-26
uni-t-ut8803e: Initial driver skeleton.

KonradTagowski:
It is more complex than uni-t-ut8802e. It is possible to imitate pressing the button using serial console etc., select, range, min/max by the specific command sent to the device. I am not sure how to implement this functionality in Sigrok

```
 case SR_CONF_SWAP:
    /* SWAP Function in this case */
        return ut8803e_send_cmd(sdi, CMD_CODE_SELECT);
    default:
```
I can't find another solution to add the functionality of the SELECT button, so I use the SWAP config key to do this. I suppose that the SELECT button will be helpful in measurement.

I reconstructed the protocol manually.


biot:
Hi,

Before I review this driver, could you please check whether it's using an already-supported DMM chip, as I mentioned in PR #220? This is usually the case for UNI-T multimeters, and supporting it may be as simple as commit f5c697bfe6d80c95051eaa1f52e1fefb54effa97.

PR #226, opened 2023-10-09
Ols read metadata in little endian

dgatf:
Ols devices are little endian. 32 bit integers for metadata values must be read in little endian (as the samples) 
If metadata is read in big endian, values for maximum rate and maximum samples are incorrect

### DIFF
--- a/bindings/python/sigrok/core/classes.i
+++ b/bindings/python/sigrok/core/classes.i
@@ -623,4 +623,13 @@ std::map<std::string, Glib::VariantBase> dict_to_map_options(PyObject *dict,
 }
 }
 
+%extend sigrok::ChannelGroup
+{
+%pythoncode
+{
+    def __repr__(self):
+        return _class_attribute_repr(self, ['name', 'channels'])
+}
+}
+
 %include "doc_end.i"


kwtaylor:
I don't think this is true? From the OLS verilog core: 

https://github.com/magnuskarlsson/ols_verilog_3.07/blob/e9c5de2271f6c555e1f9c58da65c4c3da4a6224c/meta.v#L75-L76

It sends the byte sequence for example, it sends the metadata sequence 21 00 00 60 00 for sample memory of 24K, or 0x6000, and 23 0B EB C2 00 for 200MHz sample rate, or 0x0BEBC200 -- So it's clearly big-endian.

Other SUMP-compatible devices (Bus Pirate, etc) seem to follow this as well. There is a problem in sigrok properly populating the valid sample rates from a generic SUMP device, I have created a pull request for that issue here #230

fenugrec:
It seems strange that this would suddenly be wrong for all OLS devices ?

Submitter will need to add details on what tests show this to be broken.

Sonic-Amiga:
Hello! Which metadata exactly is wrong ?
I am using arduino analyzer with OLS driver. I do experience some other problems, but metadata seems to be read correctly. At least number of channels is valid, so are supported scan rates, given by pulseview.

Sonic-Amiga:
BTW, since you also have an OLS-compatible device, can you help me to verify https://github.com/sigrokproject/libsigrok/pull/232 ? I assume that you don't experience "invalid unit size" when saving. What is your device, how many channels does it support ?

PR #228, opened 2023-11-04
[WIP] rigol-ds: add DHO800 series

stefanct:
Hello fellow DHO[89]00 owners,
I've seen multiple people asking for these devices in #sigrok on IRC. Unfortunately, I don't have time or motivation to finish the port ATM. The current patch simply adds the necessary IDs to find any DHO800 device when attached via USB - but not more than that and the protocol has changed a bit (IIRC I tested version 4 as well, the patch sets version 5).
I've tested with `LD_LIBRARY_PATH=.libs/ sigrok-cli  -d rigol-ds -l 5 --frames 1` and IIRC that simply times out. But at least it's a starting point if someone wanna beat me to it...

### DIFF
--- a/src/hardware/rigol-ds/api.c
+++ b/src/hardware/rigol-ds/api.c
@@ -185,6 +185,7 @@ enum series {
 	DS4000,
 	MSO5000,
 	MSO7000A,
+	DHO800,
 };
 
 /* short name, full name */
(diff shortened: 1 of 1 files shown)


abraxa:
Since #229 is more feature-complete than this PR, I'm closing this one. Thanks for getting the ball rolling!

PR #230, opened 2023-11-19
SUMP/OLS Enhancements

kwtaylor:
1) Properly close serial port after a failed SUMP/OLS scan, so it can be retried without restarting the application
2) Properly enumerate the valid samplerates based on returned max_samplerate metadata (rather than using hard-coded default) -- This should address this bugzilla from 2019 http://sigrok.org/bugzilla/show_bug.cgi?id=1465

Tested on linux in pulseview with Bus pirate fw 7.1 from https://github.com/BusPirate/Bus_Pirate and Arduino logic analyzer from https://github.com/gillham/logic_analyzer

### DIFF
--- a/src/hardware/openbench-logic-sniffer/api.c
+++ b/src/hardware/openbench-logic-sniffer/api.c
@@ -387,7 +387,16 @@ static int config_list(uint32_t key, GVariant **data,
 		return STD_CONFIG_LIST(key, data, sdi, cg, scanopts, drvopts,
 				       devopts);
 	case SR_CONF_SAMPLERATE:
-		*data = std_gvar_samplerates_steps(ARRAY_AND_SIZE(samplerates));
+		if (!sdi)
+			return SR_ERR_ARG;
+		devc = sdi->priv;
+		uint64_t samplerates_ovrd[3];
+		samplerates_ovrd[0] = samplerates[0];
+		samplerates_ovrd[1] = samplerates[1];
+		samplerates_ovrd[2] = samplerates[2];
+		if (devc->max_samplerate)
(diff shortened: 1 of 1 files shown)


fenugrec:
can't comment on part of this, but the "close serial port" commit is almost OK (the message is not in the proper format)

kwtaylor:
Made commit messages nicer

src/hardware/openbench-logic-sniffer/api.c
```diff
@@ -387,7 +387,16 @@ static int config_list(uint32_t key, GVariant **data,
 		return STD_CONFIG_LIST(key, data, sdi, cg, scanopts, drvopts,
 				       devopts);
 	case SR_CONF_SAMPLERATE:
-		*data = std_gvar_samplerates_steps(ARRAY_AND_SIZE(samplerates));
+		if (!sdi)
```
fenugrec, line 393: Does this func ever get called with sdi==NULL ? If so, I'm worried about the previous block (SR_CONF_DEVICE_OPTIONS) where we use sdi without checking. Could be OK, I'm not sure

abraxa: Should be ok, fx2lafw returns SR_ERR_ARG if there's no sdi, too: https://sigrok.org/gitweb/?p=libsigrok.git;a=blob;f=src/hardware/fx2lafw/api.c;h=255788bfdd0731233ca75e0a9454bb7021ce5de2;hb=HEAD#l596

Other drivers don't even support listing the sample rate and just support get/set.

abraxa:
Merged as https://sigrok.org/gitaction/libsigrok.git/5711e13..c058440, thank you!

PR #231, opened 2023-11-20
g_array_free usage

fenugrec:
Supersedes PR #207  .

only siglent-sds/protocol.c had a memory leak, but the usage in drivers.c was clarified.


fenugrec:
Merged as of c78fa9cf9416260e018c131f72e9115302deba7f

PR #232, opened 2023-11-21
openbench-logic-sniffer/protocol: Fix hardcoded unitsize

Sonic-Amiga:
Some output modules, like srzip, calculate unitsize from maximum number of
channels in the session; and throw an error if the actual datastream uses
bigger size. Remove hardcoded value of 4 and calculate it from maximum number
of channels, supported by the device. Other driver seem to do the same.

This fixes "Unexpected unit size" saving a capture, made by an OLS device with
small number of channels.

The bug was found by using gillham's Arduino analyzer
(https://github.com/gillham/logic_analyzer), which supports a maximum of 8
channels. According to the error message, srzip expected unitsize == 1.

Signed-off-by: Pavel Fedin <pavel_fedin@mail.ru>

### DIFF
--- a/src/hardware/openbench-logic-sniffer/protocol.c
+++ b/src/hardware/openbench-logic-sniffer/protocol.c
@@ -298,6 +298,11 @@ SR_PRIV int ols_get_metadata(struct sr_dev_inst *sdi)
 	/* Optionally amend received metadata, model specific quirks. */
 	ols_metadata_quirks(sdi);
 
+    /* Certain consumer modules, like srzip, don't like when we feed bigger
+	 * unitsize than they expect from maximum number of channels.
+     */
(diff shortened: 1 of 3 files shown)


Sonic-Amiga:
@gsigh I have broken down the patch into two, as you suggested; and i also reconsidered unitsize formation.

This change should keep the original behavior for devices, which support all 32 channels. I. e. it will report unitsize = 4, as previously.

I hit this problem in pulseview; i didn't recheck with cli; but i expect the same behavior.

abraxa:
Merged as https://sigrok.org/gitaction?p=libsigrok.git;a=commit;h=5711e132c324a27050c10fcc6258cd3da7bdd8fc, thank you! The logging verbosity merge was dropped as gsi already implemented an even more verbose output.

PR #234, opened 2023-11-24
Update vc96.c

schuma-gta02:
Fixes wrong detection and calculation.

### DIFF
--- a/src/dmm/vc96.c
+++ b/src/dmm/vc96.c
@@ -2,7 +2,7 @@
  * This file is part of the libsigrok project.
  *
  * Copyright (C) 2012-2013 Uwe Hermann <uwe@hermann-uwe.de>
- * Copyright (C) 2018 Matthias Schulz <matthschulz@arcor.de>
+ * Copyright (C) 2018 Matthias Schulz <matthias.schulz@mailbox.org>
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -20,6 +20,7 @@
 
 /*
  * Voltcraft 13-bytes ASCII protocol parser.
+ * derived from metex14.c
  *
  * Bytes 1-3 measuring mode, byte 4 '-' for negative,
  * bytes 5-9 value, bytes 10-11 unit, bytes 12-13 CRLF 0d 0a.
@@ -37,13 +38,11 @@
 #define LOG_PREFIX "vc96"
 
 /** Parse value from buf, byte 3-8. */
-static int parse_value(const uint8_t *buf, struct vc96_info *info,
+static int parse_value(const uint8_t *buf,
 			float *result, int *exponent)
 {
 	int i, is_ol, cnt, dot_pos;
-	char valstr[8 + 1];
-
-	(void)info;
+	char valstr[6 + 1];
 
 	/* Strip all spaces from bytes 3-8. */
 	memset(&valstr, 0, 6 + 1);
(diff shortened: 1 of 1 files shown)


harper23 (commented):
In (new) line 73-77 curly brackets are added while this style is not found elsewhere for one line statements in if-clauses. The semicolon is unnecessary.

> **!strncmp(buf, "  ", 3)**

These calls in line 84 and 96 compare a string of 2 spaces and a terminating zero with the buffer. This can be done simpler with a **strcmp(buf, "  ")**

Line 93-94 sets **is_ohm** if the buffer contains "OHM". It leaves **is_ohm** untouched if the buffer contains 2 spaces.
But line 95-96 set **is_diode** if the buffer contains "DIO". It also set **is_diode** if the buffer contains 2 spaces AND **is_volt** AND **is_milli** are set. This looks weird. Is the protocol really so strange?
Line 97-98 sets **hfe** only if the buffer contains "hfe" AND all members (is_ampere, is_volt, isresistance, is_diode) are not set. Is this really special for"hfe"?

Line 112-128 checks for units. Old order **A**, **mA**, **µA**,**V**,**mV**. New order **mA**,**µA**,**A**,**mV**,**V**. The order for check for **K** and **M** has also changed. Is this really necessary and is it better?

Line 148-150 contains unnecessary curly brackets and semicolon.

Line 159 looks like the real fix for calculation.

Line 271 changed the check for the return value of **int parse_value()**. That type of the function's return value is not of type **enum sr_error_code**. The main problem is, that the function always returns 0 (implicitly converted from SR_OK). The line 272 is never reached.

harper23:
The "cleanup curly braces" leaves an empty // comment in line 73.

The static function **parse_value()** that contains the actual fix is not called anywhere in the source code. It's dead code. The detection of ".0L" and similar strings is not present anymore. Is this intentional?

schuma-gta02:
I removed those nonreachable lines 266 and 267, they were for debugging only.
What else has to be done to get it accepted?

schuma-gta02:
Hi, another month is gone.

All request from harper23 were fixed.
What else has to done, to get this patch applied?

Matthias

src/dmm/vc96.c
```diff
@@ -63,25 +62,23 @@ static int parse_value(const uint8_t *buf, struct vc96_info *info,
 	is_ol += (!g_ascii_strcasecmp((const char *)&valstr, "-OL.")) ? 1 : 0;
 	is_ol += (!g_ascii_strcasecmp((const char *)&valstr, "-OL")) ? 1 : 0;
 	if (is_ol != 0) {
-		sr_spew("Over limit.");
```
abraxa, line 66: What's the reason this was removed?

schuma-gta02:
Hi Soeren,

those were for debugging only and are not needed anymore.

Matthias

abraxa:
okay but it's SR_SPEW log level, which is perfectly adequate for this kind of verbosity and some messages should remain for debugging when the driver isn't working as intended.

schuma-gta02:
Do You want me to put them in again?

abraxa:
I would prefer if you'd add them back in yeah, because it saves work for anyone trying to debug the driver in the future and no one runs libsigrok on SR_SPEW unless they really want to.

schuma-gta02:
Hi Soeren,

Done.
(To be honest: directly in the github Web UI.)

Matthias

abraxa:
Merged as https://sigrok.org/gitaction/libsigrok.git/6e5899e6c..e489dc1, thank you!

PR #236, opened 2024-01-17
siglent-sds: fix header size (some models), fix download speed

aksdfauytv:
also, support waveform header fragmented into multiple packets

### DIFF
--- a/src/hardware/siglent-sds/api.c
+++ b/src/hardware/siglent-sds/api.c
@@ -681,8 +681,7 @@ static int config_set(uint32_t key, GVariant *data,
 		}
 		break;
 	case SR_CONF_SAMPLERATE:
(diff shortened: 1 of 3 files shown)


gralco:
It would be good to have this be based of off my branch to avoid conflicts: https://github.com/sigrokproject/libsigrok/pull/176

Hopefully we can get both changes merged.

aksdfauytv:
> It would be good to have this be based of off my branch to avoid conflicts: #176
> 
> Hopefully we can get both changes merged.

Please, feel free to rebase your pull request on top of my changes. I believe I fixed the same issues as your comment mentions, but better :)

- I don't rely on a fixed header size (header sizes can be read from the header itself),
- I don't add any delays after ARM, just discard broken responses from Siglent,
- Some other stuff with reading the measurement.

PR #239, opened 2024-03-12
Add driver for WCH-CH569

ZhiyuanYuanNJ:
CH569 is USB3.0 super-high-speed interface microcontroller based on the RISC-V instruction set.


biot (commented):
This appears to be the fx2lafw driver with a small number of references to "WCH" added. It is in no way a working driver, and has nothing to do wth the WCH-CH569. What's the plan for this code?

abraxa:
I'm assuming there must be firmware for this MCU somewhere that matches the driver but as there is no feedback from the author, I have to close this PR unfortunately.

PR #243, opened 2024-07-30
Adalm2k

SgoSkz:
# Adalm2k
Added support for the [Adalm2k](https://www.analog.com/en/resources/evaluation-hardware-and-software/evaluation-boards-kits/adalm2000.html). 
Initial support was written but never finished #97 
Requires [libiio v1.0](https://github.com/analogdevicesinc/libiio)
## Features
- 16 Digital Inputs
- 100M sampling rate
## Missing
- Support for real time sampling
- Analog support
- Changing trigger method
## Tested
- 1kHz square wave @ 100M 10M 1M 100k 10k
- Disabling/Enabling channels
- N/A
## Bugs
- If the number of samples is too large, the buffer will fail to send
- Unknown

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -234,6 +234,12 @@ src_libdrivers_tail_la_SOURCES = src/driver_list_stop.c
 
 src_libdrivers_la_SOURCES = src/drivers.c
 
+if HW_ADALM2K_DRIVER
+src_libdrivers_la_SOURCES += \
+	src/hardware/adalm2k-driver/protocol.h \
+	src/hardware/adalm2k-driver/protocol.c \
+	src/hardware/adalm2k-driver/api.c
+endif
 if HW_AGILENT_DMM
 src_libdrivers_la_SOURCES += \
 	src/hardware/agilent-dmm/protocol.h \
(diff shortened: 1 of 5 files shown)


biot:
A few notes:
- The whitespace is all over the place in your code, looks like a mix of tabs and spaces. Also other clashing coding styles, like no space after keywords etc. Please check https://github.com/sigrokproject/libsigrok/blob/master/HACKING
- There's still lots of debug/TODO stuff in the code, to the point that it seems unfinished. It's fine to submit a driver that doesn't implement everything yet (I'd expect it, with hardware this versatile), but just leave the TODO, dead code etc stuff out.
- I'm not sure I get the dependency on libiio here. What does it bring to the table that you can't do with regular libusb calls, like every other libsigrok USB-using driver? It doesn't look to me like the kernel iio subsystem knows about the adalm2k at all, so what's libiio do for you? I'd like to avoid pulling in a libiio dependency if it's superfluous.

SgoSkz:
Alright, note taken. I will be updating the code when I have the time. 
In regards to libiio, I have no clue how to use libusb calls at this current moment (please let me know if you have any helpful resources to getting started), and it also helps with setting up the device, such as setting up buffer size, sample rate, enable/disable channels, setting up triggers, etc. 
Thank you for your feedback!

biot:
The libusb docs of course, but most of all check the code in other sigrok drivers. Lots of them use libusb to interface with the hardware.

PR #245, opened 2024-08-29
Siglent sds hd support

fredzo:
Hi,
This PR is based on the work of Eric Kuzmenko (https://github.com/sigrokproject/libsigrok/pull/176).
It's been discussed on EEVblog forum and tested by different users on SDS 2000X HD, 800X HD, 1000X HD, and 2000X plus : https://www.eevblog.com/forum/testgear/new-sigrokpulseview-hardware-support-(siglent-sds-hd-rigol-dho800-)/
I've tried not to change any behavior of the original driver.
There are known limitations on digital channels of 2000X HD for memory depth > 5Mpts due to a bug in Siglent's firmware. I am exchanging with Siglent support team about that.
Please let me know if you have any comment.
Best regard,
Frederic.


.gitignore
```diff
@@ -80,3 +80,4 @@ stamp-h?
 /tests/*.log
 /tests/*.trs
 /tests/main
+.vscode/
```
doge-gif, line 83: I'd suggest moving this to line 18, it might help accelerate the merging

src/hardware/siglent-sds/api.c
```diff
@@ -106,7 +106,7 @@ static const uint64_t timebases[][2] = {
 
 static const uint64_t vdivs[][2] = {
 	/* microvolts */
-	{ 500, 100000 },
+	{ 500, 1000000 },
```
doge-gif, line 109: Maybe we'd want to split this to another bugfix PR, although it depends on the maintainers.

src/hardware/siglent-sds/api.c
```diff
@@ -183,19 +193,39 @@ static const struct siglent_sds_vendor supported_vendors[] = {
...
+		{ 50, 1 }, { 2, 1000 }, 18, 8, 25, 1400363},
 	[SDS1000X] = {VENDOR(SIGLENT), "SDS1000X", SPO_MODEL,
-		{ 50, 1 }, { 500, 100000 }, 14, 8, 14000363},
+		{ 50, 1 }, { 500, 1000000 }, 14, 8, 25, 14000363},
 	[SDS1000XP] = {VENDOR(SIGLENT), "SDS1000X+", SPO_MODEL,
-		{ 50, 1 }, { 500, 100000 }, 14, 8, 14000363},
+		{ 50, 1 }, { 500, 1000000 }, 14, 8, 25, 14000363},
 	[SDS1000XE] = {VENDOR(SIGLENT), "SDS1000XE", ESERIES,
-		{ 50, 1 }, { 500, 100000 }, 14, 8, 14000363},
+		{ 50, 1 }, { 500, 1000000 }, 14, 8, 25, 14000363},
```
doge-gif, line 206: Same as above, maybe split into another bugfix PR

src/hardware/siglent-sds/api.c
```diff
@@ -628,17 +714,22 @@ static int config_set(uint32_t key, GVariant *data,
...
+			switch (vdivs[idx][1]) {
+			case 1:
+				cmd = g_strdup_printf("%" PRIu64 "V", p);
+				break;
+			case 1000:
+				cmd = g_strdup_printf("%" PRIu64 "MV", p);
+				break;
+			case 100000:
+				cmd = g_strdup_printf("%" PRIu64 "UV", p);
+				break;
```
doge-gif, line 731: Since we fixed the bug on uV being 10^5, maybe this should be 1,000,000 too?

doge-gif (commented):
Thanks for the PR!
May I suggest a few changes, so hopefully the merge will be faster?
(And no, I'm not a maintainer. Just another fella owning one of those scopes)

fredzo:
> Thanks for the PR! May I suggest a few changes, so hopefully the merge will be faster? (And no, I'm not a maintainer. Just another fella owning one of those scopes)

Thanks for the feedback @doge-gif ! I'll take your proposals (but I'm not creating a separate PR).
I had a few exchanges on Discord with maintainers, they're looking at it.
In the mean time, builds with this patch and the Rigol driver update are available on eevblog forum for Mac and Windows (see link in the PR text).
Best,
Frederic.

doge-gif (approved):

src/hardware/siglent-sds/protocol.c
```diff
@@ -330,110 +505,251 @@ static int siglent_sds_read_header(struct sr_dev_inst *sdi)
...
+
+						/* SDS2000X+: Since the LA sample rate is a fraction of the sample rate of the analog channels,
+							* there needs to be repeated "fake" samples inserted after each "real" sample
+							* in order to make the output match the timebase of an enabled analog channel.
+							* The scaling by a factor of 2.5 and 5 appears to be necessary due to an artifact
+							* where the instrument will present half of the entire sample quantity from the screen
+							* within a single block (625000 bytes, or 5000000 bits / samples). Which means there
+							* are some legitimate points missing that are filled with "fake" ones at larger timebases. */
+						for (int i = 0; i < samplerate_ratio-1; i++, ii++)
+							g_array_append_val(tmp_samplebuf, tmp_value);
```
swdee, line 602: I get why up-sampling of the digital data has been done here, however it has the unintended consequence of causing the Decoders to produce inaccurate data.

In my test case I have D9 which is approximately a 134kHz PWM square wave generated by MCU and CH1 which is an AC wave in sync with the PWM signal.    When no up-sampling is performed the PWM decoder correctly calculates the PWM signal on D9.

2mpts-475ms-2ms-div-ok

When a larger time base is used an up-sampling is applied to D9, the PWM decoder outputs the wrong data.

20mpts-10s-500ms-div-bad

Is it possible for some type of warning/notification to be drawn over D9 when up-sampling occurs to warn the user?

fredzo: Hi @swdee , sorry for the late reply, I must have missed the notification.
I'm not aware of any ways in Sigrok API to generate a displayable warning in acquisition process.
I understand that this might be misleading, but it's a limitation of the scope's sampling rate, not much we can do about it sorry...

src/hardware/siglent-sds/api.c
```diff
@@ -877,11 +969,15 @@ static int dev_acquisition_start(const struct sr_dev_inst *sdi)
...
+	case E11:
+		/** TODO : handle acquisition mode (sampling / average) */
+		break;
 	default:
 		break;
 	}
 
-	sr_scpi_source_add(sdi->session, scpi, G_IO_IN, 7000,
+	int tiemout = (devc->model->series->protocol == E11) ? 50 : 7000;
+	sr_scpi_source_add(sdi->session, scpi, G_IO_IN, tiemout,
```
Krakonos, line 980: Typo here `tiemout` -> `timeout`. Also, why is there such a huge difference between the E11 and all the others?

src/hardware/siglent-sds/protocol.c
Krakonos: I find this acceptable. However, I wonder if we could adjust the sampling rate so the encoders get the correct information? I expect the encoder is getting the real sampling rate, but receiving the oversampled data.

src/hardware/siglent-sds/protocol.c
```diff
@@ -261,6 +392,24 @@ SR_PRIV int siglent_sds_capture_start(const struct sr_dev_inst *sdi)
...
+		if (sr_scpi_get_int(sdi->conn, "WAV:MAXP?", &devc->max_points) != SR_OK)
+			return SR_ERR;
+		sr_dbg("Found maxp value of : %d.", devc->max_points);
+
+		/* Force trigger stop to make sure we have a consistant acquisition across channels */
+		if (siglent_sds_config_set(sdi, "TRIG:STOP") != SR_OK)
+			return SR_ERR;
+		if (sr_scpi_get_string(sdi->conn, "TRIG:STAT?", &buf) != SR_OK) {
+			g_free(buf);
+			return SR_ERR;
```
Krakonos, line 407: Should the buf contents be verified? Otherwise, I don't see the reason for calling `TRIG:STAT?`. I'm not too familiar with SCPI protocol though, so maybe this is something that needs to be called? In that case, perhaps a comment would be worth it.

src/hardware/siglent-sds/protocol.c
```diff
@@ -36,6 +36,137 @@
...
+
+/**
+ * Get the number of points of the current wave and save it to devc->num_samples.
+ *
+ * @param sdi the device instance.
+ * @param devc the device context.
+ *
+ * @return SR_ERR upon failure.
+ */
+static int siglent_get_wave_points(struct sr_dev_inst *sdi,struct dev_context *devc)
```
Krakonos, line 72: Nitpick: I don't fancy the output argument devc. When reading the function usage, it is not immediately clear what member fields are updated. Either num_samples could be passed directly (and devc would never enter the function), or the function could be renamed to reflect it updated num_samples field only (`siglent_update_num_samples`).

src/hardware/siglent-sds/protocol.c
```diff
@@ -330,110 +505,251 @@ static int siglent_sds_read_header(struct sr_dev_inst *sdi)
...
+			if (len < 0)
+				return TRUE;
+			len -= 15;
+			buffdata = g_array_sized_new(FALSE, FALSE, sizeof(uint8_t), len);
+			buf += 15; /* Skipping the data header. */
+			g_array_append_vals(buffdata, buf, len);
+			tmp_samplebuf = g_array_sized_new(FALSE, FALSE, sizeof(uint8_t), len); /* New temp buffer. */
+			for (uint64_t cur_sample_index = 0; cur_sample_index < devc->memory_depth_digital; cur_sample_index++) {
+				char sample = (char)g_array_index(buffdata, uint8_t, cur_sample_index);
+				for (int ii = 0; ii < 8; ii++, sample >>= 1) {
```
Krakonos, line 700: I suggest sample is typed as unsigned char. The shift here is implementation defined, frequently implemented as arithmetic shift. This means after 8 shifts the value would be 0 or 1, depending on the 8th bit of the original byte. Now, I realize this does not cause any issue now, since each iteration extracts the lowest bit and we don't check the remaining value.

Krakonos (commented):
I have left a few comments, but overall I find this PR of good quality. Additionally, I did some testing on SDS5034X and it seems to work, although there is one issue: the scope does not resume acquisition automatically and manually needs to be resumed before starting acquisition in pulseview. Is this on purpose to be able to trigger an even on the scope and then trasfer to the PC?

Krakonos:
@fredzo Hi! I wonder if you'd be willing to finish the work here? I posted a few comments, nothing major. Also, I wonder if you had time to PR #236 ? It seems to bring some value in parsing the headers, but I haven't spent much time with it. Personally, I'd like to typecast the data to structs to extract the fields.

I'm trying to help out with reviews and get stuff moving to upstream.

fredzo:
Hi @Krakonos ,
Thank you for taking the time to review and comment this PR!
I'm sorry but I moved to a different project (ngscopeclient) and I don't have the environment to build and patch this code anymore, nor the time to set it up :-/
I will not be able to update this PR in the near future, but feel free to take it from here and see if you finally manage to get it to upstream !...
Best,

Krakonos:
@fredzo Ok, no problem! I'll push it over the finish line, since most of the work is done and it would be a shame to let it go. I hope I'll be able to do the same for the Rigol scopes, though I don't have any on hand to test with.

I've seen the ngscopeclient on eevblog forums and might give it a try too, looks like a cool project -- the only thing I'm sad about is that all of the drivers will be reimplemented there with little hope of sharing the codebase; especially given a lof of the quirks and workarounds for different issues in the scope firmwares.

Have fun!

AJ528:
@Krakonos, do you know what the schedule is for merging this PR? I have a SDS2000X Plus and would love to see it supported.

Krakonos:
I did look into the PR a bit since and so far I wasn't able to solve the trigger issue


> ... Additionally, I did some testing on SDS5034X and it seems to work, although there is one issue: the scope does not resume acquisition automatically and manually needs to be resumed before starting acquisition in pulseview. Is this on purpose to be able to trigger an even on the scope and then trasfer to the PC?

I wonder if you could check on your device, is the behavior the same or is it the SDS5034X quirk? Or is it a desired behavior? I'm rarely using scopes with sigrok, my main usage are DMMs/PSUs and logic analyzers, so I'm not quite sure about the desire ergonomics of combining the physical controls with sigrok.

My expectation is to resume acqusition when sampling starts, then wait for trigger and sample when received.

AJ528:
I am using an SDS2354X Plus and I also see that trigger behavior. I think it could really be argued either way. On the one hand I agree that clicking "run" means the scope should restart acquisition, but I also understand the use case of "I finally captured this 1/1000 behavior and want to transfer the existing capture to pulseview for analysis". I didn't find it too clunky to sit in front of the scope, capture my desired data, then go over to pulseview to import the capture via "run". 

Regardless, I think it's worth pushing the PR through. Worst case it gets added to a list of known bugs ¯\_(ツ)_/¯

swdee:
> "I finally captured this 1/1000 behavior and want to transfer the existing capture to pulseview for analysis". I didn't find it too clunky to sit in front of the scope, capture my desired data, then go over to pulseview to import the capture via "run".

This is also how I use my scope and pulseview.

PR #250, opened 2024-10-13
Fix incorrect mappigns in the Fluke 28x unit table

andysan:
The Fluke 28x unit mapping has a couple of bugs: 'dBm' incorrectly maps to SQ_MQ_VOLTAGE and 'AAC_PLUS_DC' incorrectly maps to SR_MQ_VOLTAGE. Fix these issues and set both the AC and DC flags for sum measurements.


biot (approved):

abraxa:
Merged as https://sigrok.org/gitaction/libsigrok.git/e88b6f2..883c2ac, thank you!

PR #252, opened 2024-10-18
configure: differ between check and yes/no

Ablu:
So far one could only specify yes/no (optionally for all drivers through --enable-all-drivers={yes,no}). However, even when stating a =yes, the dependency check would override it to a =no if the dependencies are missing.

This is a bit annoying for scripts since a script may specify --enable-foo-driver=yes but in fact the build may not contain `foo-driver` due to missing build dependencies. This is especially problematic for distro's which look like they currently do not build a defined driver set - which results in random drivers missing [1].

This fixes that by following the autoconf examples [2] that allow a third state `check`. With that `no` means no, `yes` means yes (raise an error if deps are unavailable) and `check` gracefully disables drivers with unavailable dependencies.

With that --enable-all-drivers can be used to ensure that all drivers are built in. Certain drivers can still be explicitly disabled by overriding the individual driver setting.

This way distros can specify --enable-all-drivers and new drivers with unmet dependencies will cause an error rather than silently getting dropped of the build.

[1] https://sigrok.org/bugzilla/show_bug.cgi?id=1699
[2] https://www.gnu.org/software/autoconf/manual/autoconf-2.67/html_node/External-Software.html#External-Software

To: sigrok-devel@lists.sourceforge.net
Cc: Daniel Thompson <daniel.thompson@linaro.org>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -281,8 +281,8 @@ sr_check_driver_deps() {
 
 AC_ARG_ENABLE([all-drivers],
 	[AS_HELP_STRING([--enable-all-drivers],
-			[enable all drivers by default [default=yes]])],
-	[], [enable_all_drivers=yes])
+			[enable all drivers by default [default=check]])],
+	[], [enable_all_drivers=check])
 
 ## _SR_DRIVER(Device name, driver-name, var-name, [dependencies...])
 m4_define([_SR_DRIVER], [
@@ -290,9 +290,14 @@ m4_define([_SR_DRIVER], [
 		[AS_HELP_STRING([--enable-$2], [enable $1 support])],
 		[$3=$enableval], [$3=$enable_all_drivers])
 
-	AS_IF([test "x[$]$3" = xyes], [sr_hw_info=yes[]m4_ifval([$4], [
+	AS_IF([test "x[$]$3" != xno],
+		[sr_hw_info=yes[]m4_ifval([$4], [
 		sr_check_driver_deps $4 \
-			|| $3=no sr_hw_info="no (missing: $sr_deps_missing)"
+			|| if test "x[$]$3" != xcheck; then
+				AC_MSG_ERROR([driver $2 is enabled, but deps are missing: $sr_deps_missing])
+			else
+				$3=no sr_hw_info="no (missing: $sr_deps_missing)"
+			fi;
 	])], [sr_hw_info='no (disabled)'])
 	sr_driver_summary_append "$2" "$sr_hw_info"
 


abraxa:
Thanks for contributing, I your changes were merged as https://sigrok.org/gitaction/libsigrok.git/47a98d3..094c982 with some minor addition from me to make clear what the options are.

Otherwise, someone building with --enable-all-drivers may be mislead into thinking that a certain dependency is a requirement when it's actually optional.

Ablu:
Thanks a lot! LGTM!

abraxa:
Hi, I received a report that someone who's building from source suddenly has issues after this change. For him, the drivers are compiled but actually linked into the shared library. For me, however, it works fine.

Do you have any idea what could cause this?

Ablu:
Hm... Are there any details available? It seems a bit weird to me that drivers are compiled but not linked in. I did not touch that part in particular... I wonder if it is an existing bug uncovered by this being more explicit? Is it all drivers failing or just some?

Happy to help to diagnose this, but I fear more details are needed. A verbose build log would be interesting...

biot:
This still checks for the driver enable to be `yes`:
https://github.com/sigrokproject/libsigrok/blob/master/configure.ac#L304

This now being set to `check` causes `HAVE_DRIVERS` to not get set, which causes them to not be listed in `drivers.c`, so they never get added to their ELF section.

Ablu:
Hm. It is late here. Will look into this tomorrow morning!

Feel free to revert if this causes severe problems!

biot:
It seems to me that comparison on line 304 is now looking at the wrong thing. It used to be correct to see if the driver was enabled, but now that needs to look at the result of the check. Not sure where you'd find that, though.

abraxa:
Thanks for looking, Bert. Since AC_MSG_ERROR() leads to the abortion of the configure script, I'd say it's safe to check for either "yes" or "check". If "check" is set and the line has been reached, it can be assumed the script hasn't aborted and hence passed the dependency check.

abraxa:
I committed https://sigrok.org/gitaction?p=libsigrok.git;a=commit;h=f06f788118191d19fdbbb37046d3bd5cec91adb1 and it works from what I can tell. Please let me know if you disagree with the solution or issues persist.

Ablu:
@abraxa @biot thanks! Fix looks good to me! I - of course - only tested with various combinations of enable all / enable / disable... :(.

Sorry for the trouble and thanks again!

biot:
Yes, that works for me as well.

PR #253, opened 2024-11-03
support SWIG 4.2 and later

kkaempf:
otherwise building sigrok fails with
bindings/swig/enums.i:1: Error: Unknown directive '%attribute'.

Also indicate that bindings/swig/enums.i is a generated file.

### DIFF
--- a/bindings/cxx/enums.py
+++ b/bindings/cxx/enums.py
@@ -73,8 +73,10 @@ header = open(os.path.join(outdirname, 'cxx/include/libsigrokcxx/enums.hpp'), 'w
 code = open(os.path.join(outdirname, 'cxx/enums.cpp'), 'w')
 swig = open(os.path.join(outdirname, 'swig/enums.i'), 'w')
 
-for file in (header, code):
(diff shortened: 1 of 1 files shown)


kkaempf:
Not sure which version of SWIG requires this change. Tested with 4.2.1 and 4.3.0.

tofurky:
Thanks, this indeed fixed the same build error I was getting on Ubuntu 24.04 against tip of master against 4.2.0-2ubuntu1. I also found this (unanswered) from someone hitting the same issue last year: https://sourceforge.net/p/sigrok/mailman/message/51783726/

Krakonos:
I'm not sure why this stopped working, but according to the swig docs ( https://www.swig.org/Doc4.3/Library.html#Library_attributes ), the `attribute.i` should be included before the use of `%attribute`, so it's likely a coincidence it worked in the past (some other include included the `attribute.i` instead of `enums.i` ). 

@abraxa I believe this PR should be merged, it fixes compilation errors on modern systems, frequently used by developers (including me).

PR #257, opened 2024-12-07
tcp: avoid blocking in sr_fd_is_readable if using poll

Daft-Freak:
This is consistent with the select path and the function description.


abraxa:
From the poll manpage:

The timeout argument specifies the number of milliseconds that poll() should block waiting for a file descriptor to become ready.

Specifying a negative value in timeout means an infinite timeout.  Specifying a timeout of zero causes poll() to return immediately, even if no file descriptors are ready.

abraxa:
Merged as https://sigrok.org/gitaction/libsigrok.git/afc8ba4..8a77fe4, thanks!

PR #259, opened 2025-01-20
Fix data aquisition and Rohde & Schwarz support in hameg-hmo

notiflux:
Right now, the `sr_analog_init`-function is called after the data is transferred to the struct, which overwrites it with zeros again. As a result, data aquisition is not possible with this driver. This PR moves the `sr_analog_init`-call to before the data transfer.

Also, R&S oscilloscopes don't support enabling or disabling individual logic channels, some don't even support enabling or disabling PODs. I implemented a flag system that allows disabling of such capabilities per device, let me know if you think this should be solved differently though.

Finally, i changed the data output format to `MAX`, which allows to transfer the entire oscilloscope memory contents (otherwise, data is downsampled according to the horizontal zoom settings). This only works while aquisition is stopped on the device, otherwise it falls back to the default output format. This way, data aquisition is still just as fast while the scope is in RUN mode, but is also able to aquire large data blocks while it's stopped (much slower).

### DIFF
--- a/src/hardware/hameg-hmo/protocol.c
+++ b/src/hardware/hameg-hmo/protocol.c
@@ -39,7 +39,7 @@ static const char *hameg_scpi_dialect[] = {
 	[SCPI_CMD_SET_COUPLING]		      = ":CHAN%d:COUP %s",
 	[SCPI_CMD_GET_SAMPLE_RATE]	      = ":ACQ:SRAT?",
 	[SCPI_CMD_GET_ANALOG_DATA]	      = ":FORM:BORD %s;" \
-					        ":FORM REAL,32;:CHAN%d:DATA?",
+					        ":FORM REAL,32;:CHAN%d:DATA:POIN MAX;:CHAN%d:DATA?",
 	[SCPI_CMD_GET_VERTICAL_SCALE]	      = ":CHAN%d:SCAL?",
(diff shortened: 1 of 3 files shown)


src/hardware/hameg-hmo/protocol.c
```diff
@@ -1613,7 +1661,7 @@ SR_PRIV int hmo_receive_data(int fd, int revents, void *cb_data)
 	 * number of frames or after the specified number of samples, or
 	 * continue reception by starting over at the first enabled channel.
 	 */
-	if (++devc->num_frames >= devc->frame_limit || devc->num_samples >= devc->samples_limit) {
+	if ((devc->frame_limit && ++devc->num_frames >= devc->frame_limit) || devc->num_samples >= devc->samples_limit) {
```
Krakonos, line 1664: Not an issue, but I wonder what is the corner case when frame_limit needs to be checked for non-zero?

Krakonos (approved):
@abraxa I have reviewed the code and while I don't have any of the supported devices for testing, the changes are very reasonable and should not cause issues. The scope of the changes is nicely described in the PR comment and alight with the code.

src/hardware/hameg-hmo/protocol.c
notiflux: the way a frame limit of 0 is displayed in the pulseview UI is "No Limit". Without this change it would simply not aquire any data at all

PR #261, opened 2025-03-16
input: protocoldata: make sure lookup_protocol_name stays safe from NULL pointer dereferencing

fallen:
As of today, the code does not crash because all protocols have names. But the name of the first element is not checked to be non-NULL before calling strcmp(). This can result in a future crash if the code logic changes.

This is mostly a change to robustify the lookup_protocol_name() function.

Also, this commit simplifies the indexing logic of the protocols array.


fallen:
Thanks a lot @abraxa !

PR #265, opened 2025-05-13
Fix Deprecations in libzip and libcheck APIs

neuschaefer:
Several functions cause deprecation warnings when building libsigrok on a recent system. This PR aims to fix them without causing any semantic changes.

### DIFF
--- a/tests/analog.c
+++ b/tests/analog.c
@@ -393,7 +393,7 @@ START_TEST(test_analog_to_float_conv)
 		/* Convert to an array of single precision float values. */
 		ret = sr_analog_to_float(&analog, &f_out[0]);
 		if (!item->want) {
-			fail_if(ret == SR_OK,
+			ck_assert_msg(ret != SR_OK,
 				"%s: sr_analog_to_float() passed", item_text);
 			if (with_diag) {
 				fprintf(stderr, " -- expected fail, OK\n");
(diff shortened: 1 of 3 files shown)


Krakonos:
* The zip part is correct according to `libzip` documentation  at https://libzip.org/documentation/zip_add.html , the new functions are available since libzip 0.11 - release in 2013 and should be supported on all actively maintained systems.
* libcheck: This change is correct according to the documentation: https://libcheck.github.io/check/doc/check_html/check_4.html , I'm not sure when the `ck_assert_msg` etc. macros were introduces, but are available at least in 0.15.2, release 2020 -- I'd say this is safe to merge, especially since it is not required for end user to run this code. The change is semantically correct, however, introduces inconsistent indentation.

@neuschaefer Would you mind amending the PR to fix the indentation issues?

The documentation states Linux Kernel Coding Style should be used and while it does not specifically state it anywhere I can find it, tabs should be used for indents. It seems your editor changed some tabs to spaces during reformatting.

neuschaefer:
I had another look at my patches, and I think they're fine as they are: There's a difference between indentation (due to nesting, discussed in section 1 of the [Linux Kernel coding style](https://www.kernel.org/doc/html/latest/process/coding-style.html)) and visual alignment in continued lines (what the coding style calls "descendants" and covers in section 2). The coding style mentions: _"A very commonly used style is to align descendants to a function open parenthesis."_ which implies that spaces (or tabs+spaces) are used for alignment. This was already done in a few cases such as [test_samplerate](https://github.com/sigrokproject/libsigrok/pull/265/files#diff-8090656333433c85b42858cfda7094b20e94ffa0bd9136db3cda0b3926b930f0R97), although it should be noted that none of the tab+spaces alignment looks right in GitHub's diff viewer, which uses a tabstop of 4 instead the the traditional 8.

PR #268, opened 2025-05-14
raspberrypi-pico: Add link to firmware repo

neuschaefer:
Since the Raspberry Pi Pico is a programmable device with a myriad of use-cases, it needs specific firmware to implement the same protocol as the driver in src/hardware/raspberrypi-pico.

Add a link to make it easier to find the right firmware.

cc @pico-coder

### DIFF
--- a/README.devices
+++ b/README.devices
@@ -75,6 +75,11 @@ The following drivers/devices require a firmware upload upon connection:
 
      http://sigrok.org/wiki/LeCroy_LogicStudio#Firmware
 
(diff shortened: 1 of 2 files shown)


neuschaefer:
@Krakonos good idea, done

Krakonos (approved):
These changes improve documentation for the raspberrypi-pico target. I recommend these for merging.

neuschaefer:
> I would also appreciate if you could create a wiki entry for this device and driver so that the link to the firmware can be placed there, too. If you need a wiki account, let me know.

Yes, I can update the page. Please help me get a Wiki account.

PR #269, opened 2025-05-25
Add support for Korad KA3005PS

MartinJM:
This PR adds support for the Korad KA3005PS. Protocol is almost the same as for the KA3005P, except that it needs a newline[^1] at the end of every request, and sends back a newline at the end of every response. That is added in this PR.

One thing to keep in mind: This PR also sends the newline for the `*IDN?` to identify the device. I assume this will also work for the KA3005P, but I cannot check.

[^1]: It also seems to work with a carriage return, but the responses will still end with a newline. I've implemented it with newlines.

### DIFF
--- a/src/hardware/korad-kaxxxxp/protocol.c
+++ b/src/hardware/korad-kaxxxxp/protocol.c
@@ -25,12 +25,26 @@
 #define EXTRA_PROCESSING_TIME_MS  450
 
 SR_PRIV int korad_kaxxxxp_send_cmd(struct sr_serial_dev_inst *serial,
-	const char *cmd)
+	const char *cmd, bool add_newline)
 {
 	int ret;
 
-	sr_dbg("Sending '%s'.", cmd);
-	if ((ret = serial_write_blocking(serial, cmd, strlen(cmd), 0)) < 0) {
+	char* addition = "";
+	if (add_newline)
+		addition = "\n";
+
+	/* 21 was chosen here because 20 is chosen in korad_kaxxxxp_set_value */
+	char newcmd[21];
+	ret = sr_snprintf_ascii(newcmd, 21, "%s%s", cmd, addition);
+	if (ret < 0 || ret >= 21) {
+		sr_err("Error creating command: %d.", ret);
+		if (ret > 0)
+			ret = -ret; /* make errors always return negative numbers */
+		return ret;
+	}
+
+	sr_dbg("Sending '%s'.", newcmd);
+	if ((ret = serial_write_blocking(serial, newcmd, strlen(newcmd), 0)) < 0) {
 		sr_err("Error sending command: %d.", ret);
 		return ret;
 	}
@@ -71,7 +85,7 @@ SR_PRIV int korad_kaxxxxp_send_cmd(struct sr_serial_dev_inst *serial,
  * larger initial timeout period before receive data is seen.
  */
 SR_PRIV int korad_kaxxxxp_read_chars(struct sr_serial_dev_inst *serial,
-	size_t count, char *buf)
+	size_t count, char *buf, bool strip_newline)
 {
 	int timeout_first, timeout_later, timeout;
 	size_t retries_first, retries_later, retries;
@@ -146,6 +160,10 @@ SR_PRIV int korad_kaxxxxp_read_chars(struct sr_serial_dev_inst *serial,
 		timeout = timeout_later;
 		retries = retries_later;
 	}
+
+	if (strip_newline && buf[received-1] == '\n')
+		buf[--received] = 0x00;
+
 	/* TODO Escape non-printables? Seen those with status queries. */
 	sr_dbg("got %zu bytes, received: '%s'.", received, buf);
 
@@ -259,7 +277,7 @@ SR_PRIV int korad_kaxxxxp_set_value(struct sr_serial_dev_inst *serial,
 	}
 
 	if (ret == SR_OK && msg[0]) {
-		ret = korad_kaxxxxp_send_cmd(serial, msg);
+		ret = korad_kaxxxxp_send_cmd(serial, msg, devc->model->quirks & KORAD_QUIRK_NEWLINE);
 		devc->next_req_time = next_req_time(devc, TRUE, target);
 	}
 
@@ -272,7 +290,7 @@ SR_PRIV int korad_kaxxxxp_get_value(struct sr_serial_dev_inst *serial,
 	int target, struct dev_context *devc)
 {
 	int ret, count;
-	char reply[6];
+	char reply[7];
 	float *value;
 	char status_byte;
 	gboolean needs_ovp_quirk;
@@ -282,27 +300,29 @@ SR_PRIV int korad_kaxxxxp_get_value(struct sr_serial_dev_inst *serial,
 	give_device_time_to_process(devc);
 
 	value = NULL;
-	count = 5;
+	count = 6;
+
+	bool newline_quirk = devc->model->quirks & KORAD_QUIRK_NEWLINE;
 
 	switch (target) {
 	case KAXXXXP_CURRENT:
 		/* Read current from device. */
-		ret = korad_kaxxxxp_send_cmd(serial, "IOUT1?");
+		ret = korad_kaxxxxp_send_cmd(serial, "IOUT1?", newline_quirk);
 		value = &(devc->current);
 		break;
 	case KAXXXXP_CURRENT_LIMIT:
 		/* Read set current from device. */
-		ret = korad_kaxxxxp_send_cmd(serial, "ISET1?");
+		ret = korad_kaxxxxp_send_cmd(serial, "ISET1?", newline_quirk);
 		value = &(devc->current_limit);
 		break;
 	case KAXXXXP_VOLTAGE:
 		/* Read voltage from device. */
-		ret = korad_kaxxxxp_send_cmd(serial, "VOUT1?");
+		ret = korad_kaxxxxp_send_cmd(serial, "VOUT1?", newline_quirk);
 		value = &(devc->voltage);
 		break;
 	case KAXXXXP_VOLTAGE_TARGET:
 		/* Read set voltage from device. */
-		ret = korad_kaxxxxp_send_cmd(serial, "VSET1?");
+		ret = korad_kaxxxxp_send_cmd(serial, "VSET1?", newline_quirk);
 		value = &(devc->voltage_target);
 		break;
 	case KAXXXXP_STATUS:
@@ -310,8 +330,10 @@ SR_PRIV int korad_kaxxxxp_get_value(struct sr_serial_dev_inst *serial,
 	case KAXXXXP_OCP:
 	case KAXXXXP_OVP:
 		/* Read status from device. */
-		ret = korad_kaxxxxp_send_cmd(serial, "STATUS?");
+		ret = korad_kaxxxxp_send_cmd(serial, "STATUS?", newline_quirk);
 		count = 1;
+		if (newline_quirk)
+			count = 2;
 		break;
 	default:
 		sr_err("Don't know how to query %d.", target);
@@ -324,7 +346,7 @@ SR_PRIV int korad_kaxxxxp_get_value(struct sr_serial_dev_inst *serial,
 
 	devc->next_req_time = next_req_time(devc, FALSE, target);
 
-	if ((ret = korad_kaxxxxp_read_chars(serial, count, reply)) < 0) {
+	if ((ret = korad_kaxxxxp_read_chars(serial, count, reply, newline_quirk)) < 0) {
 		g_mutex_unlock(&devc->rw_mutex);
 		return ret;
 	}
--- a/src/hardware/korad-kaxxxxp/api.c
+++ b/src/hardware/korad-kaxxxxp/api.c
@@ -57,6 +57,8 @@ static const double amps_5[] = { 0, 5.1, 0.001, };
 
 static const struct korad_kaxxxxp_model models[] = {
 	/* Vendor, model name, ID reply, channels, voltage, current, quirks. */
+	{"Korad", "KA3005PS", "", 1, volts_30, amps_5,
+		KORAD_QUIRK_NEWLINE},
 	{"Korad", "KA3005P", "", 1, volts_30, amps_5,
 		KORAD_QUIRK_ID_TRAILING},
 	{"Korad", "KD3005P", "", 1, volts_30, amps_5, 0},
@@ -278,11 +280,26 @@ static GSList *scan(struct sr_dev_driver *di, GSList *options)
 		len = sizeof(reply) - 1;
 	sr_dbg("Want max %zu bytes.", len);
 
-	ret = korad_kaxxxxp_send_cmd(serial, "*IDN?");
+	/*
+	 * For the identification we send two messages due to the newline quirk
+	 * The first works on devices that require no newline to be there
+	 * The second adds support for devices that require the newline to be there
+	 * The timeout makes sure the former devices handle the cmd first
+	 */
+	ret = korad_kaxxxxp_send_cmd(serial, "*IDN?", false);
 	if (ret < 0)
 		return NULL;
 
-	ret = korad_kaxxxxp_read_chars(serial, len, reply);
+	g_usleep(100000);
+	ret = korad_kaxxxxp_send_cmd(serial, "\n", false);
+	if (ret < 0)
+		return NULL;
+
+	/*
+	 * Newline stripping enabled by default - doesn't matter for devices that
+	 * don't add one
+	 */
+	ret = korad_kaxxxxp_read_chars(serial, len, reply, true);
 	if (ret < 0)
 		return NULL;
 	sr_dbg("Received: %d, %s", ret, reply);
--- a/src/hardware/korad-kaxxxxp/protocol.h
+++ b/src/hardware/korad-kaxxxxp/protocol.h
@@ -21,6 +21,7 @@
 #ifndef LIBSIGROK_HARDWARE_KORAD_KAXXXXP_PROTOCOL_H
 #define LIBSIGROK_HARDWARE_KORAD_KAXXXXP_PROTOCOL_H
 
+#include <stdbool.h>
 #include <stdint.h>
 #include <string.h>
 #include <glib.h>
@@ -38,7 +39,8 @@ enum korad_quirks_flag {
 	KORAD_QUIRK_ID_TRAILING = 1UL << 2,
 	KORAD_QUIRK_ID_OPT_VERSION = 1UL << 3,
 	KORAD_QUIRK_SLOW_PROCESSING = 1UL << 4,
-	KORAD_QUIRK_ALL = (1UL << 5) - 1,
+	KORAD_QUIRK_NEWLINE = 1UL << 5,
+	KORAD_QUIRK_ALL = (1UL << 6) - 1,
 };
 
 /* Information on single model */
@@ -103,9 +105,9 @@ struct dev_context {
 };
 
 SR_PRIV int korad_kaxxxxp_send_cmd(struct sr_serial_dev_inst *serial,
-		const char *cmd);
+		const char *cmd, bool add_newline);
 SR_PRIV int korad_kaxxxxp_read_chars(struct sr_serial_dev_inst *serial,
-		size_t count, char *buf);
+		size_t count, char *buf, bool strip_newline);
 SR_PRIV int korad_kaxxxxp_set_value(struct sr_serial_dev_inst *serial,
 		int target, struct dev_context *devc);
 SR_PRIV int korad_kaxxxxp_get_value(struct sr_serial_dev_inst *serial,


cnf:
I think this is the same as the velleman LABPS3005DN needs i'll try and check somewhere this week on mine

MartinJM:
I'm not sure where you applied that patch? For this PR you'd need to add the `KORAD_QUIRK_NEWLINE` for your model to enable the newline sending/stripping.

I also see a change of `OUT[01]` to `OUTPUT[01]`, which would break on other models.

cnf:
Im probably doing something weird, but I am faling at applying your patch for my velleman ^^;

```
error: builder for '/nix/store/vxbwd5bik9h9vir6s196v8yp25h27bnb-libsigrok-0.5.2-unstable-2024-10-20.drv' failed with exit code 1;
        last 14 log lines:
        > Running phase: unpackPhase
        > unpacking source archive /nix/store/xcxgsps2nbp1ny2ksw0gpcs6171hqycq-libsigrok-c79a973
        > source root is libsigrok-c79a973
        > Running phase: patchPhase
        > applying patch /nix/store/vfl9vlhwsm52yxd3pripz6g9ic7fcn8y-VELLEMAN.patch
        > patching file src/hardware/korad-kaxxxxp/api.c
        > Hunk #1 FAILED at 76.
        > 1 out of 1 hunk FAILED -- saving rejects to file src/hardware/korad-kaxxxxp/api.c.rej
        > patching file src/hardware/korad-kaxxxxp/protocol.c
        > Hunk #1 FAILED at 229.
        > 1 out of 1 hunk FAILED -- saving rejects to file src/hardware/korad-kaxxxxp/protocol.c.rej
        > patching file src/hardware/korad-kaxxxxp/protocol.h
        > Hunk #1 FAILED at 40.
        > 1 out of 1 hunk FAILED -- saving rejects to file src/hardware/korad-kaxxxxp/protocol.h.rej
        For full logs, run 'nix log /nix/store/vxbwd5bik9h9vir6s196v8yp25h27bnb-libsigrok-0.5.2-unstable-2024-10-20.drv'.
```

I'll have another poke later on, my brain is a bit foggy atm i'm afraid.

MartinJM:
You'll need to apply it on top of this PR. I think you tried to apply it on master right?

I also forgot to add the newline quirk for your device - just edited it in, so you might need to grab it again.

MartinJM:
Ah I think something went wrong with my patch. I recreated it and double-checked this one, so it should work:
```patch
diff --git a/src/hardware/korad-kaxxxxp/api.c b/src/hardware/korad-kaxxxxp/api.c
index c6b9b1f6..3967a0fb 100644
--- a/src/hardware/korad-kaxxxxp/api.c
+++ b/src/hardware/korad-kaxxxxp/api.c
@@ -76,6 +76,8 @@ static const struct korad_kaxxxxp_model models[] = {
 	{"Tenma", "72-2710", "", 1, volts_30, amps_5, 0},
 	{"Velleman", "LABPS3005D", "", 1, volts_30, amps_5,
 		KORAD_QUIRK_LABPS_OVP_EN},
+	{"Velleman", "LABPS3005DN", "", 1, volts_30, amps_5,
+		KORAD_QUIRK_NEWLINE | KORAD_QUIRK_VELLEMAN},
 	{"Velleman", "PS3005D V1.3", "VELLEMANPS3005DV1.3" , 1, volts_30, amps_5,
 		KORAD_QUIRK_ID_TRAILING | KORAD_QUIRK_SLOW_PROCESSING},
 	{"Velleman", "PS3005D", "", 1, volts_30, amps_5, 0},
diff --git a/src/hardware/korad-kaxxxxp/protocol.c b/src/hardware/korad-kaxxxxp/protocol.c
index 0ce7206e..b27d8f9c 100644
--- a/src/hardware/korad-kaxxxxp/protocol.c
+++ b/src/hardware/korad-kaxxxxp/protocol.c
@@ -229,8 +229,13 @@ SR_PRIV int korad_kaxxxxp_set_value(struct sr_serial_dev_inst *serial,
 			"VSET1:%05.2f", devc->set_voltage_target);
 		break;
 	case KAXXXXP_OUTPUT:
-		sr_snprintf_ascii(msg, sizeof(msg),
-			"OUT%1d", (devc->set_output_enabled) ? 1 : 0);
+		if (devc->model->quirks & KORAD_QUIRK_VELLEMAN) {
+			sr_snprintf_ascii(msg, sizeof(msg),
+				"OUTPUT%1d", (devc->set_output_enabled) ? 1 : 0);
+		} else {
+			sr_snprintf_ascii(msg, sizeof(msg),
+				"OUT%1d", (devc->set_output_enabled) ? 1 : 0);
+		}
 		/* Set value back to recognize changes */
 		devc->output_enabled = devc->set_output_enabled;
 		break;
diff --git a/src/hardware/korad-kaxxxxp/protocol.h b/src/hardware/korad-kaxxxxp/protocol.h
index 43653d14..19aef882 100644
--- a/src/hardware/korad-kaxxxxp/protocol.h
+++ b/src/hardware/korad-kaxxxxp/protocol.h
@@ -40,7 +40,8 @@ enum korad_quirks_flag {
 	KORAD_QUIRK_ID_OPT_VERSION = 1UL << 3,
 	KORAD_QUIRK_SLOW_PROCESSING = 1UL << 4,
 	KORAD_QUIRK_NEWLINE = 1UL << 5,
-	KORAD_QUIRK_ALL = (1UL << 6) - 1,
+	KORAD_QUIRK_VELLEMAN = 1UL << 6,
+	KORAD_QUIRK_ALL = (1UL << 7) - 1,
 };
 
 /* Information on single model */
```

cnf:
hmm, not sure what I am doing wrong:

```
sr: [00:00.014127] hwdriver: sr_config_list(): key 2147418112 (NULL) sdi (nil) cg NULL -> [uint32 20000, 20001, 20003]
sr: [00:00.014149] serial: Opening serial port '/dev/ttyUSB0' (flags 1).
sr: [00:00.015854] serial: Parsing parameters from "9600/8n1".
sr: [00:00.015946] serial: Got params: rate 9600, frame 8/0/1, flow 0, rts -1, dtr -1.
sr: [00:00.015962] serial: Setting serial parameters on port /dev/ttyUSB0.
sr: [00:00.016191] serial: DBG: serial_set_params() rate 9600, 8n1
sr: [00:00.016205] serial: Flushing serial port /dev/ttyUSB0.
sr: [00:00.016227] korad-kaxxxxp: Want max 48 bytes.
sr: [00:00.016245] korad-kaxxxxp: Sending '*IDN?'.
sr: [00:00.016283] serial: Wrote 6/6 bytes.
sr: [00:00.016651] korad-kaxxxxp: want 48 bytes, timeout/retry: init 60/100, later 13/1.
sr: [00:06.030752] korad-kaxxxxp: got 0 bytes, received: ''.
sr: [00:06.030806] korad-kaxxxxp: Received: 0, 
sr: [00:06.030845] korad-kaxxxxp: Unsupported model ID '', aborting.
sr: [00:06.030882] hwdriver: Scan found 0 devices (korad-kaxxxxp).
No devices found.
```

cnf:
are you sending a newline, or the actual characters '\n' 
mine responds to the actual characterss `'n'

```
tio /dev/ttyUSB0 -b 9600
.....
[19:19:53.820] INLCRNL is set
*IDN?\nQJE3005PV1.0
STATUS?\n000
VSET1?\n00.00

```

MartinJM:
> are you sending a newline, or the actual characters '\n'

Newlines. Only noticed that you meant the actual characters just before my previous comment.

For my device it's about newlines (or carriage-returns). A literal `\\n` is very weird to me.

cnf:
bit further poking it, sending `*IDN?\\n\r` doesn't work, hell sending `*IDN?\\n ` doesn't work!

Who the hell implemented this thing, and why did I end up with one? :grin:

cnf:
```
± sigrok-cli -d korad-kaxxxxp:conn=/dev/tty-CP2102 --scan
The following devices were found:
korad-kaxxxxp:conn=/dev/tty-CP2102 - Velleman LABPS3005DN with 2 channels: V I
```
I can SET values, and enable/disable the output, but sigrok-cli can't read them back it seems:
<details>
<summary>Set / Read</summary>
```
OptiNix home-manager/patches ± sigrok-cli -d korad-kaxxxxp:conn=/dev/tty-CP2102 --config voltage_target=07.000 --set
OptiNix home-manager/patches ± sigrok-cli -d korad-kaxxxxp:conn=/dev/tty-CP2102 --show
Driver functions:
    Power supply
Scan options:
    conn
    serialcomm
    force_detect
korad-kaxxxxp:conn=/dev/tty-CP2102 - Velleman LABPS3005DN with 2 channels: V I
Supported configuration options:
    conn: /dev/tty-CP2102 (current)
    continuous: on, off
    limit_samples: 0 (current)
    limit_time: 0 (current)
    voltage: 0.000000 (current)
    voltage_target: 0.000000 (current), 31.000000, 0.010000
    current: 0.000000 (current)
    current_limit: 0.000000 (current), 5.100000, 0.001000
    enabled: on, off (current)
    regulation: CC (current)
    ocp_enabled: on (current), off
    ovp_enabled: on, off (current)
OptiNix home-manager/patches ± sigrok-cli -d korad-kaxxxxp:conn=/dev/tty-CP2102 --config current_limit=0.123 --set
OptiNix home-manager/patches ± sigrok-cli -d korad-kaxxxxp:conn=/dev/tty-CP2102 --show
Driver functions:
    Power supply
Scan options:
    conn
    serialcomm
    force_detect
korad-kaxxxxp:conn=/dev/tty-CP2102 - Velleman LABPS3005DN with 2 channels: V I
Supported configuration options:
    conn: /dev/tty-CP2102 (current)
    continuous: on, off
    limit_samples: 0 (current)
    limit_time: 0 (current)
    voltage: 0.000000 (current)
    voltage_target: 0.000000 (current), 31.000000, 0.010000
    current: 0.000000 (current)
    current_limit: 0.000000 (current), 5.100000, 0.001000
    enabled: on, off (current)
    regulation: CC (current)
    ocp_enabled: on (current), off
    ovp_enabled: on, off (current)
OptiNix home-manager/patches ±  
```
</details>

cnf:
Yeah, so control works, but reading back doesn't. from smuview i can enable the output, but it doesn't see it is on, and the voltage readings jump all over the place...

cnf:
> Hey, yeah always interested in stuff like this.
> 
> > it seems you just append 'n'
> 
> And this is giving me a good laugh, so thank you for sharing. Is your device still fine if you send `*IDN?n\n` (with the `\n` being an actual newline of course)? That should be supported by my device as well, so if that works it might be a solution for the identification. Then after the identification we know which device it is so we can use exactly what the device requires.

Yep, it doesn't seem to care about the \n, you just send the normal commands, and then 1 or 0 characters, and then an `n`, if an \n follows, it doesn't seem to notice much.
> 
> > Yeah, so control works, but reading back doesn't. from smuview i can enable the output, but it doesn't see it is on, and the voltage readings jump all over the place...
> 
> Hmmm that's too bad. Is this with just the `n`, or did you also try this with the `\n`?

either, if I run `sigrok-cli -d korad-kaxxxxp:conn=/dev/tty-CP2102 --continuous` it just gives me 0 on everything, except occastionally. I do not know why.


<details>
<summary>sigrok-cli -d korad-kaxxxxp:conn=/dev/tty-CP2102 --continuous</summary>

```
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
META regulation: CV
I: 0 mA DC
V: 0.00 V DC
META regulation: CC
I: 0 mA DC
V: 0.00 V DC
META regulation: CV
META ocp_enabled: 0
I: 0 mA DC
V: 0.00 V DC
META ocp_enabled: 1
I: 0 mA DC
V: 0.00 V DC
META ocp_enabled: 0
I: 0 mA DC
V: 0.00 V DC
META regulation: CC
META ocp_enabled: 1
I: 0 mA DC
V: 0.00 V DC
META regulation: CV
META ocp_enabled: 0
I: 0 mA DC
V: 0.00 V DC
META regulation: CC
META ocp_enabled: 1
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
META regulation: CV
I: 0 mA DC
V: 0.00 V DC
META regulation: CC
I: 0 mA DC
V: 0.00 V DC
META regulation: CV
META ocp_enabled: 0
I: 0 mA DC
V: 0.00 V DC
META ocp_enabled: 1
I: 0 mA DC
V: 0.00 V DC
META ocp_enabled: 0
I: 0 mA DC
V: 0.00 V DC
META regulation: CC
META ocp_enabled: 1
I: 0 mA DC
V: 0.00 V DC
META regulation: CV
META ocp_enabled: 0
I: 0 mA DC
V: 0.00 V DC
META regulation: CC
META ocp_enabled: 1
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
META regulation: CV
I: 0 mA DC
V: 0.00 V DC
META regulation: CC
I: 0 mA DC
V: 0.00 V DC
META regulation: CV
META ocp_enabled: 0
I: 0 mA DC
V: 0.00 V DC
META ocp_enabled: 1
I: 0 mA DC
V: 0.00 V DC
META ocp_enabled: 0
I: 0 mA DC
V: 0.00 V DC
META regulation: CC
META ocp_enabled: 1
I: 0 mA DC
V: 0.00 V DC
META regulation: CV
META ocp_enabled: 0
I: 0 mA DC
V: 0.00 V DC
META regulation: CC
META ocp_enabled: 1
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
META regulation: CV
I: 0 mA DC
V: 0.00 V DC
META regulation: CC
I: 0 mA DC
V: 0.00 V DC
META regulation: CV
META ocp_enabled: 0
I: 0 mA DC
V: 0.00 V DC
META ocp_enabled: 1
I: 0 mA DC
V: 0.00 V DC
META ocp_enabled: 0
I: 0 mA DC
V: 0.00 V DC
META regulation: CC
META ocp_enabled: 1
I: 0 mA DC
V: 0.00 V DC
META regulation: CV
META ocp_enabled: 0
I: 0 mA DC
V: 0.00 V DC
META regulation: CC
META ocp_enabled: 1
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
I: 0 mA DC
V: 0.00 V DC
META regulation: CV
I: 0 mA DC
``` 
</details>

```
OptiNix ~ $ tio /dev/tty-CP2102 -b 9600 
[18:41:52.166] tio 3.9
[18:41:52.166] Press ctrl-t q to quit
[18:41:52.168] Connected to /dev/tty-CP2102
[18:41:57.683] Switched local echo on
*IDN?nQJE3005PV1.0
VSET1?n09.00
ISET1?n0.000
ISET1:0.123nISET1?n0.123
```

src/hardware/korad-kaxxxxp/protocol.c
```diff
@@ -25,12 +25,26 @@
...
+	char newcmd[21];
+	ret = sr_snprintf_ascii(newcmd, 21, "%s%s", cmd, addition);
+	if (ret < 0 || ret >= 21) {
+		sr_err("Error creating command: %d.", ret);
+		if (ret > 0)
+			ret = -ret; // make errors always return negative numbers
+		return ret;
+	}
+
 	sr_dbg("Sending '%s'.", cmd);
```
Krakonos: This logs `cmd`, but actual command sent is `newcmd`.

src/hardware/korad-kaxxxxp/protocol.c
```diff
@@ -146,6 +160,10 @@ SR_PRIV int korad_kaxxxxp_read_chars(struct sr_serial_dev_inst *serial,
 		timeout = timeout_later;
 		retries = retries_later;
 	}
+
+	if (strip_newline && buf[received-1] == 0x0a)
```
Krakonos: I suggest to use `'\n'` instead of `0x0a`, it's more clear for people who don't have ASCII burned in their retinas :-)

MartinJM:
Hey sorry for the slow answer. I have been busy - including a move that requires me to set up my dev environment for this again.

This should still work for the Korad KA3005PS (I'll double-check when I have a dev env for this again), but I cannot verify if it still works on the other devices listed: https://github.com/sigrokproject/libsigrok/blob/c79a9738f8643fe84998d95d0578ec1035235471/src/hardware/korad-kaxxxxp/api.c#L62-L81

If we want to err on the safe side, I'd wait with merging this until someone with one of those devices tests the changes as well. On the other hand, that can take forever - but I'm not sure if that warrants testing on production.

We do know that the current state does NOT work for the Velleman LABPS3005DN, as reported by @cnf. There might still be common ground, as I believe there may still be options with this:

> Yep, it doesn't seem to care about the \n, you just send the normal commands, and then 1 or 0 characters, and then an n, if an \n follows, it doesn't seem to notice much.

But for this I can also not verify if it will break any of the other devices. They seem to be a bit more finicky than I'd like.


Thank you for review and comments, will fix/reply to them separately :)

src/hardware/korad-kaxxxxp/protocol.c
MartinJM: Yes, it would only show up as an additional newline, which I thought would not be wanted. Though on second thought I agree that it makes more sense to actually log what is being sent (including the newline).

Krakonos: Hmm, right, but still makes sense to log what is written, or maybe escape it into the string so it's clear.

Krakonos:
@MartinJM No worries, thanks for taking the time to get back to us.

I wonder if @cnf could check if the LABPS3005DN works with current master? Based on the conversation (which is pretty long and I admittedly only read the first and last third to see if it got resolved) I'm not clear if it is actually this PR that breaks it.

Also, it might be possible to scan in two phases? I suspect some devices will compare the whole "packet" received (I suspect this is USB serial?u), so if there is anything extra, it would screw things up. On the other hand, I think we can just probe the device twice and see if it responds to the other probe if the first one doesn't match. I'll check at work, we have some Korad PSUs, we might get lucky.

src/hardware/korad-kaxxxxp/protocol.c
MartinJM: It's only for debug logging though, so I'd rather not do any extra processing. I agree that it makes sense to log what is written, will change it.

For the changes, do you prefer multiple smaller commits, one larger commit containing multiple smaller changes, or a change of the original commit (including a force-push)?

-----

EDIT:

The default logger strips the newlines: https://github.com/sigrokproject/libsigrok/blob/c79a9738f8643fe84998d95d0578ec1035235471/src/log.c#L228-L243

I'll still make the change, as I understand that the logging function can be changed.

MartinJM:
> This should still work for the Korad KA3005PS (I'll double-check when I have a dev env for this again)

I tested it, and it still works perfectly fine!

> Based on the conversation I'm not clear if it is actually this PR that breaks it.

It was not supported on master either. It needs a literal `n` character at the end of the messages (where mine needs a newline: `\n`).

> Also, it might be possible to scan in two phases?

I like that idea! Will switch to that instead - marking the PR as WIP in the meantime. That should then also be a solution to the Velleman LABPS3005DN - as that can then be a third phase.

> I'll check at work, we have some Korad PSUs, we might get lucky.

Though I am still curious if it still works on the other PSUs, if you don't mind checking. I tend to be curious :)

> (which is pretty long and I admittedly only read the first and last third to see if it got resolved)

Hahaha, completely fair :)

Krakonos:
Hi! Good news, bad news.

Good news is, I got a powersupply we can test with. It' Korad KA3003P. The firmware reports `KORADKA3005PV2.0` and works on master (identifies, sets output voltage).

```
[...]
sr: [00:00.012652] serial: Opening serial port '/dev/ttyACM1' (flags 1).
sr: [00:00.015103] serial: Parsing parameters from "9600/8n1".
sr: [00:00.015147] serial: Got params: rate 9600, frame 8/0/1, flow 0, rts -1, dtr -1.
sr: [00:00.015150] serial: Setting serial parameters on port /dev/ttyACM1.
sr: [00:00.015162] serial: DBG: serial_set_params() rate 9600, 8n1
sr: [00:00.015166] serial: Flushing serial port /dev/ttyACM1.
sr: [00:00.015172] korad-kaxxxxp: Want max 48 bytes.
sr: [00:00.015177] korad-kaxxxxp: Sending '*IDN?'.
sr: [00:00.015187] serial: Wrote 5/5 bytes.
sr: [00:00.015195] korad-kaxxxxp: want 48 bytes, timeout/retry: init 60/100, later 13/1.
sr: [00:00.075268] serial: Read 17/48 bytes.
sr: [00:00.088347] korad-kaxxxxp: receive timed out, want 48, received 17.
sr: [00:00.088354] korad-kaxxxxp: got 17 bytes, received: 'KORADKA3005PV2.0'.
sr: [00:00.088358] korad-kaxxxxp: Received: 17, KORADKA3005PV2.0
sr: [00:00.088363] korad-kaxxxxp: Looking up: [KORADKA3005PV2.0].
sr: [00:00.088368] korad-kaxxxxp: Matches generic '[vendor] model [vers] [trail]' pattern.
sr: [00:00.088372] korad-kaxxxxp: Found: [Korad] [KA3005P]
sr: [00:00.088376] korad-kaxxxxp: Found: Korad KA3005P (idx 0).
[...]
```

Now, the bad news is the added newline breaks the detection:

```
[...]
sr: [00:00.012693] serial: Opening serial port '/dev/ttyACM1' (flags 1).
sr: [00:00.015128] serial: Parsing parameters from "9600/8n1".
sr: [00:00.015191] serial: Got params: rate 9600, frame 8/0/1, flow 0, rts -1, dtr -1.
sr: [00:00.015198] serial: Setting serial parameters on port /dev/ttyACM1.
sr: [00:00.015232] serial: DBG: serial_set_params() rate 9600, 8n1
sr: [00:00.015242] serial: Flushing serial port /dev/ttyACM1.
sr: [00:00.015252] korad-kaxxxxp: Want max 48 bytes.
sr: [00:00.015262] korad-kaxxxxp: Sending '*IDN?'.
sr: [00:00.015279] serial: Wrote 6/6 bytes.
sr: [00:00.015294] korad-kaxxxxp: want 48 bytes, timeout/retry: init 60/100, later 13/1.
sr: [00:06.022112] korad-kaxxxxp: got 0 bytes, received: ''.
sr: [00:06.022122] korad-kaxxxxp: Received: 0, 
sr: [00:06.022132] korad-kaxxxxp: Unsupported model ID '', aborting.
sr: [00:06.022139] hwdriver: Scan found 0 devices (korad-kaxxxxp).
sr: [00:06.022152] hwdriver: Cleaning up all drivers.
[...]
```

And I'll let it finish with some good news. I tried retrying the `*IDN?` query  first with newline, second with no newline (although I would recommend the opposite way to keep compatibility in case some other slightly different devices don't like it) and it works. The only issue I found is that timeout is currently set to 60ms and 100 retries, which results in 6s wait time. I dropped it down to 60*5 and it works, my PSU replied within 60ms (I didn't read the serial code too carefully and suspect this is not a coincidence, since the buffer is longer, the serial_read_blocking waits for the timeout time... Not ideal, though probably needed due to no delimiter appears to be sent by the PSU after the model number.

```
[...]
sr: [00:00.013357] serial: Opening serial port '/dev/ttyACM1' (flags 1).
sr: [00:00.015828] serial: Parsing parameters from "9600/8n1".
sr: [00:00.015894] serial: Got params: rate 9600, frame 8/0/1, flow 0, rts -1, dtr -1.
sr: [00:00.015900] serial: Setting serial parameters on port /dev/ttyACM1.
sr: [00:00.015919] serial: DBG: serial_set_params() rate 9600, 8n1
sr: [00:00.015924] serial: Flushing serial port /dev/ttyACM1.
sr: [00:00.015933] korad-kaxxxxp: Want max 48 bytes.
sr: [00:00.015943] korad-kaxxxxp: Sending '*IDN?'.
sr: [00:00.015957] serial: Wrote 6/6 bytes.
sr: [00:00.015970] korad-kaxxxxp: want 48 bytes, timeout/retry: init 60/5, later 13/1.
sr: [00:00.316906] korad-kaxxxxp: got 0 bytes, received: ''.
sr: [00:00.316919] korad-kaxxxxp: Received: 0, 
sr: [00:00.316928] korad-kaxxxxp: Sending '*IDN?'.
sr: [00:00.316953] serial: Wrote 5/5 bytes.
sr: [00:00.316969] korad-kaxxxxp: want 48 bytes, timeout/retry: init 60/5, later 13/1.
sr: [00:00.377054] serial: Read 17/48 bytes.
sr: [00:00.390143] korad-kaxxxxp: receive timed out, want 48, received 17.
sr: [00:00.390154] korad-kaxxxxp: got 17 bytes, received: 'KORADKA3005PV2.0'.
sr: [00:00.390162] korad-kaxxxxp: Received: 17, KORADKA3005PV2.0
sr: [00:00.390174] korad-kaxxxxp: Looking up: [KORADKA3005PV2.0].
sr: [00:00.390187] korad-kaxxxxp: Matches generic '[vendor] model [vers] [trail]' pattern.
sr: [00:00.390194] korad-kaxxxxp: Found: [Korad] [KA3005P]
sr: [00:00.390201] korad-kaxxxxp: Found: Korad KA3005P (idx 1).
[...]
```

Krakonos:
@MartinJM I also noticed the PR #219 seems to do a similar thing, including the two phase scan. I originally though we could add the PN here too, but it seems to require more stuff there. I suggest we merge this PR first, since it's more up to date, we can then update the other one just to modify the protocol part (if the original author is still interested/somebody has the device for testing).

MartinJM:
> Good news is, I got a powersupply we can test with

Cool, thank you for testing! Yeah a bit sad it didn't work out of the box like this, but good to know the other options.

> I also noticed the PR https://github.com/sigrokproject/libsigrok/pull/219 seems to do a similar thing

Oh good find, not sure how I missed that before writing/opening this...

> including the two phase scan

I think they just send `*IDN?` first, and if they don't get a response, then send another `\n`. So not a full second scan. Interesting idea though, which may work fine - even when also considering the Velleman DN (`*IDN?` + `n\n` should work I think). I'll add that in first as it should be a simpler change than a full second (and potentially even third) scan.

-----

EDIT:

> The only issue I found is that timeout is currently set to 60ms and 100 retries, which results in 6s wait time. I dropped it down to 60*5 and it works

This is a good point that I hadn't considered yet. Would it work fine if we just send `*IDN?n\n*IDN?` in a single message? It works for mine, I think it will work for the Velleman DN, so if it also works for yours then that may be the fastest yet complete solution?

MartinJM:
@Krakonos I pushed a version that uses `*IDN?\n*IDN?` to identify the hardware. Works fine on mine, if you still have the Korad KA3003P available, can you please test if that also works fine on that? If it works it would avoid having to wait for a timeout for all devices.

Krakonos:
@MartinJM Good attempt, but unfortunately, it doesn't work. The PSU does not reply unless I give at least some delay there. It likely processes a command after some period of inactivity. I have zero idea how somebody thought this is a good idea.

In any case, this works:

```
diff --git a/src/hardware/korad-kaxxxxp/api.c b/src/hardware/korad-kaxxxxp/api.c
index 7e4f2eb7..515970af 100644
--- a/src/hardware/korad-kaxxxxp/api.c
+++ b/src/hardware/korad-kaxxxxp/api.c
@@ -287,7 +287,9 @@ static GSList *scan(struct sr_dev_driver *di, GSList *options)
         * But the normal ones need an `*IDN?` without anything at the end.
         * So this sends a combination that satisfies these constraints.
         */
-       ret = korad_kaxxxxp_send_cmd(serial, "*IDN?\n*IDN?", false);
+       ret = korad_kaxxxxp_send_cmd(serial, "*IDN?\n", false);
+        g_usleep(40000);
+       ret = korad_kaxxxxp_send_cmd(serial, "*IDN?", false);
        if (ret < 0)
                return NULL;
```

The value 40000 is the minimum I found working, 30000 already fails. I'd suggest using higher delay  Not the most elegant, but might be an alternative. This device is not automatically scanned based on vid/pid (it uses a generic converter), so I don't think it is an issue.

I wonder if the following patch would work with your PSU?

```
diff --git a/src/hardware/korad-kaxxxxp/api.c b/src/hardware/korad-kaxxxxp/api.c
index 7e4f2eb7..1836af55 100644
--- a/src/hardware/korad-kaxxxxp/api.c
+++ b/src/hardware/korad-kaxxxxp/api.c
@@ -287,7 +287,9 @@ static GSList *scan(struct sr_dev_driver *di, GSList *options)
         * But the normal ones need an `*IDN?` without anything at the end.
         * So this sends a combination that satisfies these constraints.
         */
-       ret = korad_kaxxxxp_send_cmd(serial, "*IDN?\n*IDN?", false);
+       ret = korad_kaxxxxp_send_cmd(serial, "*IDN?", false);
+        g_usleep(40000);
+       ret = korad_kaxxxxp_send_cmd(serial, "\n", false);
        if (ret < 0)
                return NULL;
```

MartinJM:
> Good attempt, but unfortunately, it doesn't work.

Cool, thank you for trying!


> I wonder if the following patch would work with your PSU?

Yeah the second patch works perfectly fine. The length of the sleep between the two doesn't seem to matter too much. I tried it with up to 10 seconds (`g_usleep(10000000);`) and it still works fine.

It also works as the following, which could also enable detecting the Velleman DN - if @cnf is still available for testing:

```diff
diff --git a/src/hardware/korad-kaxxxxp/api.c b/src/hardware/korad-kaxxxxp/api.c
index 7e4f2eb7..9b26ce7a 100644
--- a/src/hardware/korad-kaxxxxp/api.c
+++ b/src/hardware/korad-kaxxxxp/api.c
@@ -287,7 +287,9 @@ static GSList *scan(struct sr_dev_driver *di, GSList *options)
         * But the normal ones need an `*IDN?` without anything at the end.
         * So this sends a combination that satisfies these constraints.
         */
-       ret = korad_kaxxxxp_send_cmd(serial, "*IDN?\n*IDN?", false);
+       ret = korad_kaxxxxp_send_cmd(serial, "*IDN?", false);
+       g_usleep(500000);
+       ret = korad_kaxxxxp_send_cmd(serial, "n\n", false);
        if (ret < 0)
                return NULL;
```

The current state of this PR actually doesn't work too well for my device. When closing the connection it seems to restart the serial connection, which is not ideal. With your patch it doesn't do that, so it's even better for my device as well!

Krakonos:
Ok, I think we can call that a decent solution then. If @cnf will get in touch to test some of the variants we can tune it a little more. I'm not sure about the `"n\n"`, didn't it require the backslash character? If so, I'd say it should be `"\\n\n"`. In any case. my PSU doesn't care, since it evaluates the command right after the timeout.

On other note, could you please rebase this on top of master and clean up the commits (squash the iterative commits and update commit messages to prefix with "drivername: "? Once this is done, I'd be happy to recommend this for merging.

Krakonos:
Also:

> The current state of this PR actually doesn't work too well for my device. When closing the connection it seems to restart the serial connection, which is not ideal. With your patch it doesn't do that, so it's even better for my device as well!

I don't see how these changes could affect what happens on the device after the connection closes. But these PSUs are weird.

The only thing I see on mine is it beeps after some inactivity (30s? give or take), but does not seem reset any state.

MartinJM:
> no \n no /n no \\n .... just a single n...
https://github.com/sigrokproject/libsigrok/pull/269#issuecomment-3289656544

Should just be a single `n`. Attached is a more complete patch for the Velleman based on the earlier work, but I think it makes sense putting that in a new PR at this point.
[0001-Add-Velleman-DN-support.patch](https://github.com/user-attachments/files/23824054/0001-Add-Velleman-DN-support.patch)

> On other note, could you please rebase this on top of master and clean up the commits (squash the iterative commits and update commit messages to prefix with "drivername: "? Once this is done, I'd be happy to recommend this for merging.

Will do! I'll remove the draft status when complete.

> I don't see how these changes could affect what happens on the device after the connection closes. But these PSUs are weird.

Yeah I'm not sure either. But as long as this version works I'm fine with it :)

Krakonos:
@MartinJM Great, thanks! I gave it a quick test and all seems good.

@cnf In case you're interested, please try the patch posted above to see if it resolves your issue as well. Tag us here/reach out on the mailing list if so, we'll open and merge a new PR.

@abraxa Please check this PR out, I believe it is good to go for merging.

PS: I wonder if it's possible to flash/cross flash those PSUs to some common decent firmware. These are not the only quirks, I recall some people having more serious functionality related issues...

Krakonos (approved):

cnf:
> @MartinJM Great, thanks! I gave it a quick test and all seems good.
> 
> @cnf In case you're interested, please try the patch posted above to see if it resolves your issue as well. Tag us here/reach out on the mailing list if so, we'll open and merge a new PR.
> 
> @abraxa Please check this PR out, I believe it is good to go for merging.
> 
> PS: I wonder if it's possible to flash/cross flash those PSUs to some common decent firmware. These are not the only quirks, I recall some people having more serious functionality related issues...

Sorry, only just say the messages... let me see what to apply where, and give it a test...

edit: i'm struggling a bit figuring out what patches to apply.

So i am using master, adding https://patch-diff.githubusercontent.com/raw/sigrokproject/libsigrok/pull/269.patch on top of that, and then 
```
diff --git a/src/hardware/korad-kaxxxxp/api.c b/src/hardware/korad-kaxxxxp/api.c
index 7e4f2eb7..9b26ce7a 100644
--- a/src/hardware/korad-kaxxxxp/api.c
+++ b/src/hardware/korad-kaxxxxp/api.c
@@ -287,7 +287,9 @@ static GSList *scan(struct sr_dev_driver *di, GSList *options)
         * But the normal ones need an `*IDN?` without anything at the end.
         * So this sends a combination that satisfies these constraints.
         */
-       ret = korad_kaxxxxp_send_cmd(serial, "*IDN?\n*IDN?", false);
+       ret = korad_kaxxxxp_send_cmd(serial, "*IDN?", false);
+       g_usleep(500000);
+       ret = korad_kaxxxxp_send_cmd(serial, "n\n", false);
        if (ret < 0)
                return NULL;
```

Is that right? It is failing to apply for me.

cnf:
> @MartinJM Great, thanks! I gave it a quick test and all seems good.
> 
> @cnf In case you're interested, please try the patch posted above to see if it resolves your issue as well. Tag us here/reach out on the mailing list if so, we'll open and merge a new PR.
> 
> @abraxa Please check this PR out, I believe it is good to go for merging.
> 
> PS: I wonder if it's possible to flash/cross flash those PSUs to some common decent firmware. These are not the only quirks, I recall some people having more serious functionality related issues...

Yes, a not crappy common firmware would be awesome :P I don't know if these are flashable, or if any code is available for them?

MartinJM:
> edit: i'm struggling a bit figuring out what patches to apply.

The patch linked above ( https://github.com/user-attachments/files/23824054/0001-Add-Velleman-DN-support.patch ) on top of this branch.


> > PS: I wonder if it's possible to flash/cross flash those PSUs to some common decent firmware. These are not the only quirks, I recall some people having more serious functionality related issues...
> 
> Yes, a not crappy common firmware would be awesome :P I don't know if these are flashable, or if any code is available for them?

Could be nice if possible indeed. Though supporting devices out of the box is also good imo.

cnf:
> > edit: i'm struggling a bit figuring out what patches to apply.
> 
> The patch linked above ( https://github.com/user-attachments/files/23824054/0001-Add-Velleman-DN-support.patch ) on top of this branch.
ok, so i am, and I am getitng this error:

```
        > src/hardware/korad-kaxxxxp/protocol.c: In function 'korad_kaxxxxp_set_value':
┃        > src/hardware/korad-kaxxxxp/protocol.c:241:53: error: 'struct dev_context' has no member named 'set_output_enables'; did you mean 'set_output_enabled'?
┃        >   241 |                                 "OUTPUT%1d", (devc->set_output_enables) ? 1 : 0);
┃        >       |                                                     ^~~~~~~~~~~~~~~~~~
┃        >       |                                                     set_output_enabled
┃        >   CC       src/hardware/lecroy-logicstudio/api.lo
```

i guess it was renamed somewhere, and i'm patching against a wrong version somewhere...

MartinJM:
Oh no that's a typo in my patch - my bad. You can change `set_output_enables` to `set_output_enabled` to solve that.

cnf:
> Oh no that's a typo in my patch - my bad. You can change `set_output_enables` to `set_output_enabled` to solve that.

yeah, I just figured that out :P 

```
hydra ~ $ sigrok-cli -d korad-kaxxxxp:conn=/dev/tty-CP2102 --scan
The following devices were found:
korad-kaxxxxp:conn=/dev/tty-CP2102 - Velleman LABPS3005DN with 2 channels: V I
hydra ~ $ sigrok-cli -d korad-kaxxxxp:conn=/dev/tty-CP2102 --show
Driver functions:
    Power supply
Scan options:
    conn
    serialcomm
    force_detect
sr: korad-kaxxxxp: Unsupported model ID '', aborting.
No devices found.
zsh: exit 1     sigrok-cli -d korad-kaxxxxp:conn=/dev/tty-CP2102 --show
```

```
sigrok-cli --version
sigrok-cli 0.8.0

Libraries and features:
- libsigrok 0.6.0/4:0:0 (rt: 0.6.0/4:0:0).
 - Libs:
  - glib 2.84.4 (rt: 2.84.4/8404:4)
  - zlib 1.3.1
  - libzip 1.11.4
  - minilzo 2.10
  - libserialport 0.1.1/1:0:1 (rt: 0.1.1/1:0:1)
  - libusb-1.0 1.0.29.11953 API 0x0100010b
  - hidapi 0.15.0
  - bluez 5.83
  - libftdi 1.5
  - Host: x86_64-pc-linux-gnu, little-endian.
  - SCPI backends: TCP, serial, USBTMC.
- libsigrokdecode 0.6.0/4:0:0 (rt: 0.6.0/4:0:0).
 - Libs:
  - glib 2.84.4 (rt: 2.84.4/8404:4)
  - Python 3.13.7 / 0x30d07f0 (API 1013, ABI 3)
  - Host: x86_64-pc-linux-gnu, little-endian.
```

Krakonos:
@cnf Hi! Would you mind running this with `-l 5` ? This looks like you got some response, but the model ID is empty string. I wonder what happened there? However, I suspect we might need to modify the patch:
```
ret = korad_kaxxxxp_send_cmd(serial, "n\n", false);
```

... I suggest to try in "\\n\n", or maybe we'll need to split it off. It'd be nice if you could play with this code a little bit and figure out what gives you a response and what doesn't. Unfortunately, there are too many combinations that need to be tried out and I'm starting to think we might not be able to auto detect all of them easily.

cnf:
@Krakonos sure!

```
hydra ~ $ sigrok-cli -d korad-kaxxxxp:conn=/dev/tty-CP2102 --scan -l 5
sr: [00:00.000444] log: libsigrok loglevel set to 5.
sr: [00:00.000455] backend: libsigrok 0.6.0/4:0:0.
sr: [00:00.000478] backend: Libs: glib 2.84.4 (rt: 2.84.4/8404:4), zlib 1.3.1, libzip 1.11.4, minilzo 2.10, libserialport 0.1.1/1:0:1 (rt: 0.1.1/1:0:1), libusb-1.0 1.0.29.11953 API 0x0100010b, hidapi 0.15.0, bluez 5.83, libftdi 1.5.
sr: [00:00.000489] backend: Host: x86_64-pc-linux-gnu, little-endian.
sr: [00:00.000496] backend: SCPI backends: TCP, serial, USBTMC.
sr: [00:00.000509] backend: Firmware search paths:
sr: [00:00.000528] backend:  - /home/cnf/.local/share/sigrok-firmware
sr: [00:00.000533] backend:  - /nix/store/ryz4705fca0hszl5vrjrmyhlppwd61xp-libsigrok-unreleased/share/sigrok-firmware
sr: [00:00.000541] backend:  - /usr/share/sigrok-firmware
sr: [00:00.000575] backend:  - /var/lib/flatpak/exports/share/sigrok-firmware
sr: [00:00.000591] backend:  - /home/cnf/.local/share/flatpak/exports/share/sigrok-firmware
sr: [00:00.000608] backend:  - /nix/store/cad0ndrfahnbls1fzrr5kmpf3cvk9ypk-network-manager-applet-1.36.0/share/sigrok-firmware
sr: [00:00.000619] backend:  - /nix/store/nmgh4491415yj8v3pa9kaxgfz5kh1maf-gnome-mimeapps/share/sigrok-firmware
sr: [00:00.000625] backend:  - /nix/store/k4s8sd32q946mgddp7p0rlwral44b6c4-desktops/share/sigrok-firmware
sr: [00:00.000635] backend:  - /home/cnf/.local/share/flatpak/exports/share/sigrok-firmware
sr: [00:00.000646] backend:  - /var/lib/flatpak/exports/share/sigrok-firmware
sr: [00:00.000655] backend:  - /home/cnf/.nix-profile/share/sigrok-firmware
sr: [00:00.000666] backend:  - /nix/profile/share/sigrok-firmware
sr: [00:00.000673] backend:  - /home/cnf/.local/state/nix/profile/share/sigrok-firmware
sr: [00:00.000693] backend:  - /etc/profiles/per-user/cnf/share/sigrok-firmware
sr: [00:00.000699] backend:  - /nix/var/nix/profiles/default/share/sigrok-firmware
sr: [00:00.000713] backend:  - /run/current-system/sw/share/sigrok-firmware
sr: [00:00.000722] backend:  - /nix/store/n6npyz7c05f2bwhf8yidc1z0xnksc6vm-gnome-shell-48.2/share/gsettings-schemas/gnome-shell-48.2/sigrok-firmware
sr: [00:00.001668] backend: Sanity-checking all drivers.
sr: [00:00.001687] backend: Sanity-checking all input modules.
sr: [00:00.001698] backend: Sanity-checking all output modules.
sr: [00:00.002211] backend: Sanity-checking all transform modules.
srd: libsigrokdecode loglevel set to 5.
sr: [00:00.013934] hwdriver: sr_config_list(): key 2147418112 (NULL) sdi (nil) cg NULL -> [uint32 20000, 20001, 20003]
sr: [00:00.014217] serial: Opening serial port '/dev/tty-CP2102' (flags 1).
sr: [00:00.038023] serial: Parsing parameters from "9600/8n1".
sr: [00:00.038200] serial: Got params: rate 9600, frame 8/0/1, flow 0, rts -1, dtr -1.
sr: [00:00.038214] serial: Setting serial parameters on port /dev/tty-CP2102.
sr: [00:00.049261] serial: DBG: serial_set_params() rate 9600, 8n1
sr: [00:00.049336] serial: Flushing serial port /dev/tty-CP2102.
sr: [00:00.049380] korad-kaxxxxp: Want max 48 bytes.
sr: [00:00.049466] korad-kaxxxxp: Sending '*IDN?'.
sr: [00:00.049537] serial: Wrote 5/5 bytes.
sr: [00:00.149678] korad-kaxxxxp: Sending 'n'.
sr: [00:00.149738] serial: Wrote 2/2 bytes.
sr: [00:00.154163] korad-kaxxxxp: want 48 bytes, timeout/retry: init 60/100, later 13/1.
sr: [00:06.169078] korad-kaxxxxp: got 0 bytes, received: ''.
sr: [00:06.169099] korad-kaxxxxp: Received: 0, 
sr: [00:06.169110] korad-kaxxxxp: Unsupported model ID '', aborting.
sr: [00:06.169123] hwdriver: Scan found 0 devices (korad-kaxxxxp).
sr: [00:06.169145] hwdriver: Cleaning up all drivers.
hydra ~ $                                    
```

though it doesn't respond to scan anymore, not sure what changed

I have tried various combinations (see https://github.com/sigrokproject/libsigrok/pull/269#issuecomment-3316110594) it seems to need `<command>(0 or 1 char)n` but i can play around some tomorrow.

cnf:
[LABPS3005DN.patch](https://github.com/user-attachments/files/24015876/LABPS3005DN.patch)
ok, so this one pretty reliable works for scan, and show.

of course, it still won't detect my actually set values... maybe this one just needs its own driver? or the unified firmware :P

Krakonos:
@cnf Hi! I finally found time to retest your patch. Unfortunately, it breaks KA3003P and possibly others. Since there are more issues with your model, I believe we should just drop it for the time being. If you want to get it finished, I see two ways:

1. Make it work with forced detection by keeping the scan algorithm as is and adding the quirk as you proposed.

```
$ sigrok-cli -d korad-kaxxxxp:conn=/dev/ttyACM0:force_detect=LABPS3005DN --show
```

2. Create a new driver if you find it's too different or hacky. Perhaps with a limited functionality.

... I think taking the first approach is going to be somewhat easier. However, I suggest you open a new PR for this work and work out other parameters as well -- it doesn't make sense to merge the device support if other commands don't work.

Krakonos:
@abraxa I recommend this PR for merging.

It adds new device support, while keeping compatibility with (at least some) of the old devices.

PR #275, opened 2025-10-13
Add Sipeed SLogic Analyzer Driver Support

taonyx:
This pull request adds support for the Sipeed SLogic series logic analyzers ([SLogic Combo8](http://wiki.sipeed.com/slogic), [SLogic16U3](http://wiki.sipeed.com/slogic16u3)) to libsigrok. Key changes include:

- Integration of a new driver for Sipeed SLogic analyzers, with new source files: src/hardware/sipeed-slogic-analyzer/{api.c, protocol.c, protocol.h}.
- Addition of udev rules for Sipeed devices in contrib/60-libsigrok.rules (vendor ID 0x359f, product IDs 0x0300 and 0x3031).
- Driver supports device detection, configuration, sampling rates, channel selection, and acquisition control for SLogic Combo 8 and SLogic16U3 models.

This contribution enables libsigrok to work with Sipeed SLogic logic analyzers, broadening hardware compatibility.

PR #277, opened 2025-12-14
std: add std_i32_idx()

Mixaill:
spinoff from #275, needed for convenient operations with `SR_CONF_NUM_LOGIC_CHANNELS` which is `SR_T_INT32`


Krakonos:
@Mixaill Thanks for the contribution. I wonder if we should also add u32 and i64? Although not used currently, I find it more convenient to have these functions predefined for the common types. One would expect u32 to exist when i32 exists and vice versa.

@abraxa Any opinion on this?

Otherwise, this is OK to merge -- though I'd be happier to see all the types included.

PR #278, opened 2025-12-25
Master

martinboller:
Added support for Rigol MSO5504 as some owners of MSO5000 series has changed to the model designation to enable 500ps horizontal scale selection.


martinboller (commented):
Added support for RIGOL MSO5504 providing 500ps horizontal scale with a firmware hack

martinboller (commented):
Added support for RIGOL MSO5504 with 500 ps horicontal scale.

PR #280, opened 2026-01-18
Adjust Python bindings for SWIG >= 4.4.

aryler:
Function signature changed in SWIG 4.4. Details:
https://github.com/numpy/numpy/commit/1e424dae42a2d560520b6e053e8e60ac4205bfc7

Could maybe use NULL instead of nullptr; they should be the same with SWIG < 4.4.
Minor detail. Not sure what's better.


Krakonos:
Hi!

Let's say the API change is a bit unfortunate. Based on the commit mentioned, the 0 is correct. However, I believe the previous version should use `NULL` instead of `nullptr`, since `nullptr` is a C23 feature ( https://en.cppreference.com/w/c/language/nullptr.html ). It might be mistaked for `nullptr` in C++ that was introduced much earlier, but the code in question is C.

aryler:
Hi, thanks for looking into this.

Here, you can see similar, using NULL. They haven't removed the Python 2 guard (yet?).
https://github.com/coin3d/pivy/pull/139/commits/b8e55e20a0ea3bef1a039216a6fd2b73854b817f

Krakonos:
Hi! Yes, that's correct. But the PR as submitted uses `nullptr` (which was inherited from the original code, but I think it's worth fixing as well).

Krakonos (approved):
The change is correct. The only reservation I have is to fix the `nullptr` -> `NULL`, but that was not the scope of this PR.

johnbeard:
Based on the change that introduced this in Swig 4.4, I think it's expected to return -1 on failure rather than 0: https://github.com/swig/swig/commit/6c3bc2d18a864a2515d4abe59ba4f9c2ff36f84f

Because the value will be used in this if:

```
  if (SWIG_mod_exec(m) != 0)
  {
    Py_DECREF(m);
    return;
  }
```

PR #283, opened 2026-02-17
rdtech-rd: Fix <inf> samples for RD6006

DidierA:
On my RD6006, no current or voltage value where available: they all returned `inf`.
I tracked the bug to the current and voltage multipliers being 0, because they were not initialized.
This was due to the test in `rdtech_dps_update_range` for multiple ranges, which returns false in this case, causing the function to return without calling `rdtech_dps_update_multipliers`.

With this patch, the multipliers are initialized, and both sigrok-cli and smuview obtain the correct values.

The `devc->model->n_ranges <= 1` test had been introduced with PR #205 to support RD6012P.  I do not own any other RD6xxx power supply, so I could only test this with RD6006.

Example run:
Before this fix:
```
$ sigrok-cli -d rdtech-rd:conn=/dev/ttyUSB1 --samples 1
FRAME-BEGIN
V: inf V DC
I: inf mA DC
P: 11.93 W
FRAME-END
```

After this fix:
```
$ sigrok-cli -d rdtech-rd:conn=/dev/ttyUSB1 --samples 1
FRAME-BEGIN
V: 17.27 V DC
I: 692 mA DC
P: 11.95 W
FRAME-END
```


DidierA:
@mdjurfeldt I would appreciate feedback on this, does this fix seem appropriate? Can you test on your RD6012P if it does not break ?
Than you.

mdjurfeldt:
@DidierA Hi, looking at the code, I don't think it should break. I can try to test it during the coming few days.

PR #287, opened 2026-03-24
Fix SWIG 4.4 Python return type

johnbeard:
In SWIG 4.4, there is a multi-step init process. The %init snippet now needs to return an int -1 to satify the return type of SWIG_mod_exec.

Previously, it was a null pointer.

----

The multi-step change is mentioned in https://github.com/swig/swig/blob/master/RELEASENOTES under `SWIG-4.4.0`


johnbeard:
Sorry, I didn't see https://github.com/sigrokproject/libsigrok/pull/280 was the same thing.

PR #297, opened 2026-07-06
Pxview fork

haikumuse:


haikumuse:
hello there, this request has some private info code, and i want to delete it, can you help me?

Krakonos:
I'm not sure it's possible to fully remove a PR. I think your best bet is to delete the branch you used to open the PR, potentially taking the whole repo private. Since the PR is not merged, it only shows the state of your repository. I would:

1. Reset pxview-fork on your repo to libsigrok master. This will make this PR show empty diff.
2. Take measures to remove your private code form your public fork (or make the fork private on gh).

Note that while I didn't check the contents of the PR and don't know the exact reason. However, it is usually best to publish the supporting code under an open license, unless there is a legal reason not to (contains 3rd party or licensed code). Companies come and go, opensource stays. Having opensource support will also give you a wider audience. This holds especially true if you're already working on a private fork.

(no that the project is that active, I've been trying to push for some revival, but with little success so far)

